PR #25, opened 2014-08-14
Remove trailing slash in URL if it exists

Procta:
Added regular expression to remove trailing slash if included in URL. Api::__construct

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -72,6 +72,9 @@ class Api
         AuthenticationInterface $authentication,
         ClientInterface $client = null
     ) {
+        //Regular expression to remove trailing slash
+        $endpoint = preg_replace('{/$}', '', $endpoint);
+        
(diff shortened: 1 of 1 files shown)


src/Jira/Api.php
```diff
@@ -72,6 +72,9 @@ public function __construct(
         AuthenticationInterface $authentication,
         ClientInterface $client = null
     ) {
+        //Regular expression to remove trailing slash
+        $endpoint = preg_replace('{/$}', '', $endpoint);
```
aik099: This can be simplified like this and comment won't be needed anymore:

```php
$endpoint = rtrim($endpoint, '/');
```

jpastoor:
Expanded in #67.

aik099:
Thank you @Procta  . To expedite the 2.0.0 release the @jpastoor would be making necessary changes based upon your commits and therefore this PR will be closed.

PR #26, opened 2014-08-29
Add expand parameter to getCreateMeta()

arnested:
The [createmeta method](https://developer.atlassian.com/static/rest/jira/6.0.html#id164910) accepts an [expand parameter](https://developer.atlassian.com/static/rest/jira/6.0.html#expansion).

This is needed if I want to expand fields in the response.


jpastoor:
Looks solid

aik099:
Sorry for the delay. Merging, thank you @arnested .

PR #29, opened 2014-09-19
Some features added

Dewstar:
Added Delete in CurlClient
Added function to resolve an issue in Api
Added function to reopen an issue in Api
Added function to delete watchers in Api
Added function to delete comments in Api
Added function to delete attachments in Api

### DIFF
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -82,6 +82,8 @@ class CurlClient implements ClientInterface
             } else {
                 curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($data));
             }
+        } else if($method == "DELETE") {
+            curl_setopt($curl, CURLOPT_CUSTOMREQUEST, "DELETE");
         } else {
             if ($method == "PUT") {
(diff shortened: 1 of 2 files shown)


aik099:
Not sure if it's a good idea to hardcode any transition-specific actions in general API class. Since in JIRA workflow and transition names can be customized then for some users such call might end up with error.

Better to have universal transition method, that would accept:
* transition name
* fields to change

aik099:
I've changed my mind. Please rebase this and it can be merged. We can later refactor to remove duplicate transition finding code code and make actual transition names customizable.

aik099:

src/Jira/Api/Client/CurlClient.php
```diff
@@ -82,6 +82,8 @@ public function sendRequest($method, $url, $data = array(), $endpoint, Authentic
             } else {
                 curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($data));
             }
+        } else if($method == "DELETE") {
```
aik099: This is already addressed in #76.

jpastoor:
Planning on rebasing this myself if original author does not want to after #137 is finished.

PR #32, opened 2014-10-07
Allow anonymous access in "CurlClient" class

digitalkaoz:
public jiras allow anonymous access...so there is no need to disallow it here

### DIFF
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -26,6 +26,7 @@ namespace chobie\Jira\Api\Client;
 
 use chobie\Jira\Api\Authentication\AuthenticationInterface;
 use chobie\Jira\Api\Authentication\Basic;
+use chobie\Jira\Api\Authentication\Anonymous;
 use chobie\Jira\Api\Exception;
 use chobie\Jira\Api\UnauthorizedException;
 
@@ -58,7 +59,7 @@ class CurlClient implements ClientInterface
         $isFile = false,
         $debug = false
     ) {
-        if (!($credential instanceof Basic)) {
+        if (!($credential instanceof Basic) && !($credential instanceof Anonymous)) {
             throw new \Exception(sprintf('CurlClient does not support %s authentication.', get_class($credential)));
         }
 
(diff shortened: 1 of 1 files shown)


src/Jira/Api/Client/CurlClient.php
```diff
@@ -52,10 +52,6 @@ public function __construct()
      */
     public function sendRequest($method, $url, $data = array(), $endpoint, AuthenticationInterface $credential, $isFile = false, $debug = false)
     {
-        if (!($credential instanceof Basic)) {
```
aik099: Instead of removing use approach taken in `PHPClient`:

```php
if (!($credential instanceof Basic) && !($credential instanceof Anonymous)) {
    throw new \Exception(sprintf("PHPClient does not support %s authentication.", get_class($credential)));
}
```

src/Jira/Api/Client/CurlClient.php
```diff
@@ -65,7 +61,9 @@ public function sendRequest($method, $url, $data = array(), $endpoint, Authentic
         curl_setopt($curl, CURLOPT_URL, $endpoint . $url);
         curl_setopt($curl, CURLOPT_HEADER, 0);
         curl_setopt($curl, CURLOPT_RETURNTRANSFER, 1);
-        curl_setopt($curl, CURLOPT_USERPWD, sprintf("%s:%s", $credential->getId(), $credential->getPassword()));
+        if ($credential instanceof Basic) {
```
aik099: Please replace with `if (!($credential instanceof Anonymous)) {` (approach from `PHPClient` class) to allow custom authentication ways to be used.

digitalkaoz:
done

src/Jira/Api/Client/CurlClient.php
```diff
@@ -52,10 +53,10 @@ public function __construct()
      */
     public function sendRequest($method, $url, $data = array(), $endpoint, AuthenticationInterface $credential, $isFile = false, $debug = false)
     {
-        if (!($credential instanceof Basic)) {
-            throw new \Exception(sprintf("CurlClient does not support %s authentication.", get_class($credential)));
+        if (!($credential instanceof Basic) && !($credential instanceof Anonymous)) {
+            throw new \Exception(sprintf("PHPClient does not support %s authentication.", get_class($credential)));
```
aik099: I haven't noticed, that exception message was containing class name. Please change `PHPClient` back to `CurlClient`.

src/Jira/Api/Client/CurlClient.php
```diff
@@ -52,10 +53,10 @@ public function __construct()
...
         }
-
+
```
aik099: Please revert this change. Also enabling setting in IDE to trim whitespaces on empty lines is a recommended approach.

digitalkaoz: did it completel in the github ui, so sorry

aik099: It really makes contributions easier indeed. Don't worry, we'll be adding automated coding standard checks later on and then you can see check result right no Scrutinizer CI website.

aik099:
Looks good. Please also rebase.

digitalkaoz:
so all done @aik099 also squashed everything

aik099:
Looks good. Thank you @digitalkaoz , merging.

digitalkaoz:
your welcome :)

PR #34, opened 2014-12-11
Update edit_issue.php

tot-ra:


aik099:
Looks good. I'm not sure if we're actually supporting `stdClass` objects in the API. Probably PHP just converts them into array down the road.

PR #35, opened 2015-01-12
Support avatarId key in IssueType, throw Exception if response is not a JSON

testfairy:
Recent changes in Atlassian broke jira-api-restclient.
1. avatarId is now returned in issueTypes (jira-api-restclient threw Exception upon unrecognised key.)
2. Accidentally using http://*.atlassian.net instead of https:// will redirect and use GET instead of POST.
3. Using GET when POST is required would have returned an HTTP error, but CurlClient returned it as a string. Thus, Api.php failed to parse.

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -28,6 +28,7 @@ use chobie\Jira\Api\Authentication\AuthenticationInterface;
 use chobie\Jira\Api\Client\ClientInterface;
 use chobie\Jira\Api\Result;
 use chobie\Jira\Api\Client\CurlClient;
+use chobie\Jira\Api\Exception;
 
 class Api
 {
@@ -469,6 +470,10 @@ class Api
 
         if (strlen($result)) {
             $json = json_decode($result, true);
+            if (!is_array($json)) {
+                 throw new Exception("JIRA Rest server returns unexpected result: " . $result);
+            }
+
             if ($this->options & self::AUTOMAP_FIELDS) {
                 if (isset($json['issues'])) {
                     if (!count($this->fields)) {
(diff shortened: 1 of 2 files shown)


src/Jira/Api.php
```diff
@@ -469,6 +470,10 @@ public function api(
 
         if (strlen($result)) {
             $json = json_decode($result, true);
+            if (!is_array($json)) {
```
aik099: Better to check for `null` specifically as this clearly indicates the error. Not sure if that won't backfire if some of API calls really has returns null.

src/Jira/IssueType.php
```diff
@@ -45,6 +45,7 @@ class IssueType
         "iconUrl",
         "name",
         "subtask",
+        "avatarId",
```
aik099: Please revert changes to this file, because avatar support was added in #50.

aik099:
> avatarId is now returned in issueTypes (jira-api-restclient threw Exception upon unrecognised key.)

Solved as part of #50.

> Accidentally using http://*.atlassian.net instead of https:// will redirect and use GET instead of POST.

That is not the problem, that API client needs to be solving. Also JIRA doesn't force you to use HTTPS.

> Using GET when POST is required would have returned an HTTP error, but CurlClient returned it as a string. Thus, Api.php failed to parse.

That is not the problem, that API client needs to be solving. Whoever is using the library must supply valid input. Just guessing from the call result if it's failed or not isn't bulletproof solution.

@jpastoor , overall I think this PR needs to be closed.

aik099:

aik099:
Closing, because it's the developer's responsibility to specify the correct Jira API URL (both hostname & protocol).

PR #39, opened 2015-03-02
fix the file attachment for curl

ElectricMaxxx:
I i have fixed the file attachment with curl, by adding the right string to the `$data` array. When it exists the string is created by the `curl_file_create`  function, as a string otherwise.

I have put this transformation into the curl client, as i don't know if the other clients are using a different way of transfering file data.

I have no clue about your tests atm, will fix my PR when any issue occurs.


aik099:
Thank you. This was fixed in other way in #52 . If it doesn't work for you feel free to create another issue.

PR #42, opened 2015-04-02
fix typo in function name

josevh:
fix typo

getPriorties() should be getPriorities()

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -300,7 +300,7 @@ class Api
      *
      * @return mixed
      */
-    public function getPriorties()
+    public function getPriorities()
     {
         if (!count($this->priorities)) {
             $priorities = array();


jpastoor:
BC breaking but good change.

src/Jira/Api.php
```diff
@@ -300,7 +300,7 @@ public function getVersions($projectKey)
      *
      * @return mixed
      */
-    public function getPriorties()
+    public function getPriorities()
```
aik099: Please also add `getPriorties` (old method name) that would:

* trigger `E_USER_DEPRECATED` notice (lets people know, that new method should be called instead)
* call `getPriorities` (new method name) internally (to prevent BC break)

jpastoor:
Expanded upon in #68

aik099:
Thank you @josevh . To expedite the 2.0.0 release the @jpastoor would be making necessary changes based upon your commits and therefore this PR will be closed.

PR #43, opened 2015-04-14
Add method to create remote links for an issue

elmi82:
This method allows the user to add remote links to an issue. Remote link
point to systems other than JIRA itself, e.g. a web site or a wiki
page.

Creating links between JIRA issues accomplished by a different API and
thus beyond the scope of this method.


aik099:
Ready to be merged. Thank you @elmi82 .

Issue #24, opened 2014-08-12
Example to CREATE an issue


saada:
Hey there!
Is there an example to help create an issue?
I saw the edit issue example, but no create example.

Thanks

GeRiY:
Hey there :)
This option is in the master branch?
I try to use edit issue example but I get all time this error: 
Field 'Status' cannot be set. It is not on the appropriate screen, or unknown.

Someone can write to me a little tutorial for how to use edit issue?
I try with issue fields keys.

Thx
Gergo

aik099:
@GeRiY , here is a discussion about a similar error on the Atlassian forum: https://community.atlassian.com/t5/Jira-questions/quot-Field-customfield-10012-cannot-be-set-It-is-not-on-the/qaq-p/664236 .

You have to add the `Status` field to the corresponding issue editing screen in the JIRA configuration.

PR #44, opened 2015-05-18
Add an example of issue creation

t-keller:
As requested in the issue : https://github.com/chobie/jira-api-restclient/issues/24

### DIFF
--- a/examples/create_issue.php
+++ b/examples/create_issue.php
@@ -1,0 +1,7 @@
+<?php
+require dirname(__FILE__) ."/common.php";
+
+$api = getApiClient();
+$api->createIssue('<Projet Key>', '<Summary>', '<Task Type Id>', array(
+	'description' => '<Issue Description>'
+));


src/Jira/Api/Client/CurlClient.php
```diff
@@ -78,6 +78,7 @@ public function sendRequest($method, $url, $data = array(), $endpoint, Authentic
         if ($method == "POST") {
             curl_setopt($curl, CURLOPT_POST, 1);
             if ($isFile) {
+                curl_setopt($curl, CURLOPT_SAFE_UPLOAD, false);
```
aik099: This would result in an error prior PHP 5.5, where this option was added.

This was fixed in #52 .

jpastoor:
PR contains commits that should be no part of this, like changing the composer file and such. Feel free to resubmit with only the example commit.

README.md
```diff
@@ -1,40 +1,10 @@
-# Jira Api Rest Client
+# Jira Api Rest Client (Forked)
```
aik099: Please revert changes to this file.

composer.json
```diff
@@ -1,6 +1,6 @@
 {
```
aik099: Please revert changes to this file.

composer.json
```diff
@@ -1,6 +1,6 @@
...
-    "name": "chobie/jira-api-restclient",
-    "description": "JIRA REST API.",
+    "name": "t-keller/jira-api-restclient",
```
aik099: To use your fork instead of original package you need to specify this in your `composer.json` file:

```json
"repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/t-keller/jira-api-restclient"
        }
    ],
```

aik099:

PR #46, opened 2015-06-24
Add expand options for getProjects and getProject

umbertix:


src/Jira/Api.php
```diff
@@ -157,14 +157,14 @@ public function getAttachment($attachmentId)
         return $result;
     }
 
-    public function getProjects()
+    public function getProjects($expand = '')
     {
-        return $this->api(self::REQUEST_GET, "/rest/api/2/project");
+        return $this->api(self::REQUEST_GET, "/rest/api/2/project",array('expand' => $expand), true);
```
aik099: The `, true` is setting `$return_as_json` to `true` and therefore making return value just an array instead of `Result` class. This is a BC break.

Would you please explain why you did this?

src/Jira/Api.php
```diff
@@ -157,14 +157,14 @@ public function getAttachment($attachmentId)
         return $result;
     }
 
-    public function getProjects()
+    public function getProjects($expand = '')
```
aik099: I would rather make `$expand` parameter and array and then concatenate values by comma before making API call (right in this method).

src/Jira/Api.php
```diff
@@ -157,14 +157,14 @@ public function getAttachment($attachmentId)
...
     {
-        return $this->api(self::REQUEST_GET, "/rest/api/2/project");
+        return $this->api(self::REQUEST_GET, "/rest/api/2/project",array('expand' => $expand), true);
     }
 
-    public function getProject($projectKey)
+    public function getProject($projectKey, $expand = '')
```
aik099: I would rather make `$expand` parameter and array and then concatenate values by comma before making API call (right in this method).

aik099:

aik099:
@umbertix , you might have deleted branch and pushed changes and that caused PR to be auto-closed. When rebasing there is no need to delete the branch.

PR #47, opened 2015-06-25
Fix create attachments for PHP 5.5

DerMika:
I was having problems with the createAttachment() using PHP 5.6. Turns out there is a new way to handle file uploads in Curl from PHP 5.5 and above: the function curl_file_create().

I also provided a way to specify the file name you want to use in Jira for the attachment you're uploading. It shouldn't break existing code, it's an optional parameter. If unused the filename defaults to basename() of the file you provide.

And I improved on some PHPDocs. Sorry, it's a fetish of mine...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 vendor
 composer.lock
+.idea/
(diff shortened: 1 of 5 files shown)


aik099:
Thank you.

* The main problem (inability to upload files on PHP 5.5) using `CurlClient` was solved in #52 .
* The code style fixes will be made as part of #62 .

If that doesn't work for you, please open another issue.

PR #50, opened 2015-09-09
Added an additional field of avatarId to IssueTypes

addersuk:
Added an additional field of avatarId to IssueTypes as Jira Cloud is now returning this field via the api and causing the call to fail.

I've added a test to show this is now working.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,11 @@
             "chobie\\" : "src"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "chobie\\Tests\\": "tests"
+        }
+    },
     "extra": {
         "branch-alias": {
             "dev-master": "2.0-dev"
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -23,6 +23,11 @@
             "chobie\\" : "src"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "chobie\\Tests\\": "tests"
```
aik099: Nice. While you're at it please namespace and rename `Jira_Api_Authentication_BasicTest` class.

aik099:
Please squash.

aik099:
Actually this one be merged and we'll fix existing tests separately.

aik099:
The `avatarId` parameter was added to issue types in JIRA 6.4.

PR #51, opened 2015-09-24
Fix path in uploaded filename

shmaltorhbooks:
if uploaded file located not in current directory it has sloppy name with path in JIRA: "/var/www/path/file.txt" instead of "file.txt"

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -433,7 +433,7 @@ class Api
     {
         $options = array_merge(
             array(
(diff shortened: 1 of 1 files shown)


aik099:
Sorry for the delay. 

Are you using `CurlClient` or `PHPClient`? If you're using `PHPClient`, then problem was solved in #59.

aik099:
This was fixed in #87 . The issue was only with `CurlClient`.

PR #52, opened 2015-09-24
Fix broken files upload on php > 5.6.0

shmaltorhbooks:
CURLOPT_SAFE_UPLOAD - TRUE to disable support for the @ prefix for uploading files in CURLOPT_POSTFIELDS. 

Added in PHP 5.5.0 with FALSE as the default value. PHP 5.6.0 changes the default value to TRUE.

http://php.net/manual/ru/function.curl-setopt.php

### DIFF
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -71,7 +71,9 @@ class CurlClient implements ClientInterface
         curl_setopt($curl, CURLOPT_SSL_VERIFYHOST, 0);
         curl_setopt($curl, CURLOPT_VERBOSE, $debug);
         if ($isFile) {
-            curl_setopt($curl, CURLOPT_SAFE_UPLOAD, false);
+            if (defined('CURLOPT_SAFE_UPLOAD')) {
+                curl_setopt($curl, CURLOPT_SAFE_UPLOAD, false);
+            }
             curl_setopt($curl, CURLOPT_HTTPHEADER, array('X-Atlassian-Token: nocheck'));
         } else {
             curl_setopt($curl, CURLOPT_HTTPHEADER, array("Content-Type: application/json;charset=UTF-8"));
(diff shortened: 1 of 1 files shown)


shmaltorhbooks:
ping

aik099:
Sorry for the delay, @shmaltorhbooks . I suppose you're trying to upload a file with `@` in it's name and this doesn't work?

aik099:
According to https://bugs.php.net/bug.php?id=66677 enabling this option by default would also wrong security-wise because anyone that can pass POST parameters can upload local files on the system through it.

src/Jira/Api/Client/CurlClient.php
```diff
@@ -71,6 +71,9 @@ public function sendRequest($method, $url, $data = array(), $endpoint, Authentic
         curl_setopt($curl, CURLOPT_SSL_VERIFYHOST, 0);
         curl_setopt($curl, CURLOPT_VERBOSE, $debug);
         if ($isFile) {
+            if (defined('CURLOPT_SAFE_UPLOAD')) {
+                curl_setopt($curl, CURLOPT_SAFE_UPLOAD, false);
```
aik099: I'm confused. Documentation for this option says this:

```
TRUE to disable support for the @ prefix for uploading files in CURLOPT_POSTFIELDS, which means that values starting with @ can be safely passed as fields. CURLFile may be used for uploads instead.
```

so the value `false` is enabling unsafe mode, that allows `@` usage for filename uploads.

aik099: Ah, I see now. The `createAttachment` method uses `@` for file upload according to https://docs.atlassian.com/jira/REST/latest/#api/2/issue/{issueIdOrKey}/attachments-addAttachment .

Then it's OK for now, but might be bad in long term from security point of view when an attacker has control over name of the filename, that is uploaded it can trick CURL into uploading any other expect specified file.

PR #53, opened 2015-10-30
Added deleteIssue function and changed curlClient to support DELETE r…

marek-knappe:
…equest

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -150,6 +150,24 @@ class Api
     }
 
 
+     /**
+     * Delete issue
+     *
+     * @param $issueKey should be YOURPROJ-221
+     * @param $deleteSubtasks if all subtask should be deleted
+     * @return mixed
+     */
+    public function deleteIssue($issueKey, $deleteSubtasks = 'true')
+    {
+        return $this->api(
+            self::REQUEST_DELETE, sprintf("/rest/api/2/issue/%s", $issueKey), 
+            array (
+                'deleteSubtasks' => $deleteSubtasks
+                )
+        );
+    }   
+
+
     public function getAttachment($attachmentId)
     {
         $result = $this->api(self::REQUEST_GET, "/rest/api/2/attachment/$attachmentId", array(), true);
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -82,12 +82,14 @@ class CurlClient implements ClientInterface
             } else {
                 curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($data));
             }
-        } else {
-            if ($method == "PUT") {
-                curl_setopt($curl, CURLOPT_CUSTOMREQUEST, "PUT");
-                curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($data));
-            }
+        } elseif ($method == "DELETE") {
+            curl_setopt($curl, CURLOPT_CUSTOMREQUEST, "DELETE");
+
+        } elseif ($method == "PUT") {
+            curl_setopt($curl, CURLOPT_CUSTOMREQUEST, "PUT");
+            curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($data));
         }
+        
 
         $data = curl_exec($curl);
 


src/Jira/Api.php
```diff
@@ -72,6 +72,9 @@ public function __construct(
         AuthenticationInterface $authentication,
         ClientInterface $client = null
     ) {
+        //Regular expression to remove trailing slash
+        $endpoint = preg_replace('{/$}', '', $endpoint);
```
aik099: By replace code with solution below no explanatory comment is needed:

```php
$endpoint = rtrim($endpoint, '/');
```

Sounds like useful change, but not sure how it's related to this PR.

src/Jira/Api.php
```diff
@@ -238,6 +259,20 @@ public function addComment($issueKey, $params)
...
+    /**
+     * get all worklogs for an issue
+     *
+     * issue key should be YOURPROJ-22
+     *
+     * @param $issueKey
+     * @param $params
+     * @return mixed
+     */
+    public function getWorklogs($issueKey, $params)
```
aik099: Not related to this PR. Please submit as separate PR.

aik099:
@marek-knappe I see now why you have all your PR code mixed together. You're not using feature branches and submitting PR's from `master` branch where all is added.

Unfortunately GitHub doesn't allow to change PR branch and therefore the only viable approach is this:

1. we close the PR
2. on your repo clone you create a feature branch (e.g. `delete-issue-feat`) and put new code in there
3. push it to GitHub
3. submit new PR from pushed branch

src/Jira/Api.php
```diff
@@ -150,6 +153,24 @@ public function editIssue($issueKey, $params)
...
+     * @param $issueKey should be YOURPROJ-221
+     * @param $deleteSubtasks if all subtask should be deleted
+     * @return mixed
+     */
+    public function deleteIssue($issueKey, $deleteSubtasks = 'true')
+    {
+        return $this->api(
+            self::REQUEST_DELETE, sprintf("/rest/api/2/issue/%s", $issueKey), 
+            array (
+                'deleteSubtasks' => $deleteSubtasks
```
aik099: Please change `$deleteSubtasks` into `$deleteSubtasks ? 'true' : 'false'`.

src/Jira/Api.php
```diff
@@ -150,6 +153,24 @@ public function editIssue($issueKey, $params)
...
 
 
+     /**
+     * Delete issue
+     *
+     * @param $issueKey should be YOURPROJ-221
+     * @param $deleteSubtasks if all subtask should be deleted
+     * @return mixed
+     */
+    public function deleteIssue($issueKey, $deleteSubtasks = 'true')
```
aik099: The trick that JIRA needs boolean specified as strings is it's internal thing and doesn't need to be exposed via API client interface. This way please replace `'true'` with `true`.

src/Jira/Api.php
aik099: Actually just remove it from this PR, because there is another PR doing the same thing: #25 .

marek-knappe:
@aik099 
The reason why #53 and #57 are submitted together is because if you would choose to just merge #57 you wouldn't have #53 changes and it wouldn't work :) 
anyway, will fix your recommendation and submit another PR

aik099:
> The reason why #53 and #57 are submitted together is because if you would choose to just merge #57 you wouldn't have #53 changes and it wouldn't work :) 
anyway, will fix your recommendation and submit another PR

In such case I usually submit different PRs and mention somewhere that another PR needs to be merged first for this one to work.

marek-knappe:
If you find it better (less going back and forward) I'm happy for you or anyone to make it properly :)

src/Jira/Api/Client/CurlClient.php
```diff
@@ -82,12 +82,14 @@ public function sendRequest($method, $url, $data = array(), $endpoint, Authentic
             } else {
                 curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($data));
             }
-        } else {
-            if ($method == "PUT") {
-                curl_setopt($curl, CURLOPT_CUSTOMREQUEST, "PUT");
-                curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($data));
-            }
+        } elseif ($method == "DELETE") {
+            curl_setopt($curl, CURLOPT_CUSTOMREQUEST, "DELETE");
```
aik099: The DELETE calls can have parameters. The correct implementation is done in #76.

PR #58, opened 2015-12-27
Merges 2015

jpastoor:
Not sure if this is still being looked at, but for convenience sake if chobie returns:
I merged the open pull requests (except some duplicates or stuff I was not sure about) and merged in in a big pull request.


jpastoor:
Pull requests will be evaluated again on a case by case basis.

PR #59, opened 2016-02-25
Use basename for file uploads

ubermuda:
There's a problem when using absolute paths for file uploads.

The upload seems to work fine, but then when you want to view the image in JIRA you get an error message:

```
Ouch! We can't load the image.

/secure/attachment/10909/%2Fvagrant%2Fweb%2Fattachments%2Fe7f29917-811e-41cd-8fed-6bd485e47afe.jpeg
```

### DIFF
--- a/src/Jira/Api/Client/PHPClient.php
+++ b/src/Jira/Api/Client/PHPClient.php
@@ -99,7 +99,7 @@ class PHPClient implements ClientInterface
                 $header[] = 'Content-Type: multipart/form-data; boundary=' . $boundary;
 
                 $__data = "--" . $boundary . "\r\n" .
-                    "Content-Disposition: form-data; name=\"file\"; filename=\"" . $filename . "\"\r\n" .
+                    "Content-Disposition: form-data; name=\"file\"; filename=\"" . basename($filename) . "\"\r\n" .
                     "Content-Type: application/octet-stream\r\n\r\n" .
                     file_get_contents($filename) . "\r\n";
                 $__data .= "--" . $boundary . "--\r\n";


aik099:
You're absolutely correct @ubermuda . The `filename` part of `Content-Disposition` header should be filename only without path.

jpastoor:
Hi @ubermuda,

If you like you can send this pull request as well to https://github.com/jpastoor/jira-api-restclient.git and I'll add it to the list of fixes since the original developer is not maintaining the lib anymore.

Kind regards,
Joost Pastoor

aik099:
@jpastoor , is there any way of contacting original developer to add you as maintainer (e.g. sending e-mail to chobieeee@php.net address, which is mentioned in `composer.json`)?

Having another repo where all is merged which is registered under different package name on GitHub isn't too good. Even if your repo has same package name people need to specify `repositories` key in their `composer.json` to use it.

jpastoor:
Yeah thats what I tried couple months ago but he was not responding, sadly. (Although he is active on github on other projects)

aik099:
I've sent e-mail now as well with suggestion to:

1. move repo to organization to have better permission control over it
2. grant access to Packagist package
3. grant access to repo in that organization to us

aik099:
@ubermuda if you'll be using `CurlClient` instead of `PHPClient` there all would work fine, right?

ubermuda:
Hey guys! Sorry I got a bit under the water.

So, seeing as you now have commit rights here, should I still move this PR to the other repo or do we stay here?

> @ubermuda if you'll be using CurlClient instead of PHPClient there all would work fine, right?

File upload with CurlClient does not work at all (JIRA returns a 500), I don't know why

aik099:
> So, seeing as you now have commit rights here, should I still move this PR to the other repo or do we stay here?

No need to submit to other repo, because me @jpastoor were granted maintainer permissions and can merge PRs.

> File upload with CurlClient does not work at all (JIRA returns a 500), I don't know why

Just a guess, but maybe PUT method used to handle upload isn't handled in `CurlClient` at all. I recommend at least reporting it as issue (or maybe separate PR if you know how to fix it).

aik099:
Merging, thanks @ubermuda .

ubermuda:
> Just a guess, but maybe PUT method used to handle upload isn't handled in CurlClient at all. I recommend at least reporting it as issue (or maybe separate PR if you know how to fix it).

I'll look into it and keep you updated! Thanks.

PR #66, opened 2016-02-27
Bunch of CS fixes

aik099:
We don't have any tests and that's why changes, that are of 2 types:

* not affecting code at all, e.g. whitespaces
* made by automatic clever fixing tools (e.g. PHP_CodeSniffer), e.g. replacing double with single quotes

What's not fixed in there is DocBlocks. I guess they deserve separate PR.


jpastoor:
I think you can rebase & merge this :+1:

aik099:
Ah, it was good to merge until some recent PRs get merged in. I'll do the rebase then.

aik099:
Rebase version pushed. Ready for review.

aik099:
Turned out, that one PRs were using TABs for indentation instead of SPACEs. I've fixed that as part of my PR as well.

PR #67, opened 2016-02-28
Remove trailing slash in URL if it exists

jpastoor:
Update of #25.

- Moved logic from constructor to setEndpoint
- Changed regex to rtrim
- Wrote unit test
- Rebased on current master

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -72,9 +72,6 @@ class Api
         AuthenticationInterface $authentication,
         ClientInterface $client = null
     ) {
-        //Regular expression to remove trailing slash
-        $endpoint = preg_replace('{/$}', '', $endpoint);
-        
         $this->setEndPoint($endpoint);
         $this->authentication = $authentication;
 
@@ -101,13 +98,16 @@ class Api
     }
 
     /**
-     * set end point url.
+     * Set Endpoint URL
      *
-     * @param $url
+     * @param string $url
      */
     public function setEndPoint($url)
     {
         $this->fields = array();
+
+        // Remove trailing slash in the url
+        $url = rtrim($url, '/');
 
         $this->endpoint = $url;
     }
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -72,6 +72,9 @@ class Api
         AuthenticationInterface $authentication,
         ClientInterface $client = null
     ) {
+        //Regular expression to remove trailing slash
+        $endpoint = preg_replace('{/$}', '', $endpoint);
+        
         $this->setEndPoint($endpoint);
         $this->authentication = $authentication;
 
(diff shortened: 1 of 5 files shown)


src/Jira/Api.php
```diff
@@ -106,6 +106,9 @@ public function setEndPoint($url)
     {
         $this->fields = array();
 
+        //Regular expression to remove trailing slash
```
aik099: Please change the following:

1. add a space after `//`
2. add a dot at the end of the comment text

tests/Jira/ApiTest.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Tests\chobie\Jira;
```
aik099: Are you sure these tests are auto-loaded? The namespace for tests in `autoload-dev` part of `composer.json` says `chobie\\Tests\\` namespace.

Probably need to patch it in there.

tests/Jira/ApiTest.php
```diff
@@ -0,0 +1,23 @@
...
+ * Class ApiTest
+ *
+ * @package Tests\chobie\Jira
+ */
+class ApiTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * Tests that any trailing slash in the endpoint url is removed before being stored in the object state
+     */
+    public function testSetEndpointTrailingSlash()
```
aik099: While you're at it please also add a test that shows that url is kept as is when no trailing `/` is used.

aik099:
Today’s review completed. When done making changes please add inline comment in each of them.

tests/Jira/ApiTest.php
jpastoor: Done

tests/Jira/ApiTest.php
jpastoor: Done 00451d3

src/Jira/Api.php
jpastoor: Done 1, I'm no fan of 2

tests/Jira/Api/Authentication/BasicTest.php
```diff
@@ -1,13 +1,17 @@
 <?php
 
-class Jira_Api_Authentication_BasicTest extends PHPUnit_Framework_TestCase
+namespace Tests\chobie\Jira;
+
+use chobie\Jira\Api\Authentication\Basic;
+
+class Jira_Api_Authentication_BasicTest extends \PHPUnit_Framework_TestCase
```
aik099: Since you namespace test classes why not rename `Jira_Api_Authentication_BasicTest` into `BasicTest` to make autoloading work?

aik099:
And you can safely squash when you're done fixing.

tests/Jira/Api/Authentication/BasicTest.php
jpastoor: Fixed

aik099:
Merging, thanks @jpastoor .

PR #68, opened 2016-02-28
Fixed typo in Api::getPriorities()

jpastoor:
Extension on #42 
- As by @aik099 comment, reinstated the old method name with a deprecation warning.


aik099:
Merging. Thanks @jpastoor .

PR #71, opened 2016-02-28
Added .idea to .gitignore to hide PHPStorm generated project files

jpastoor:
Since PHPStorm is the most widely used IDE it makes sense to add .idea to gitignore so the generated files don't clutter the git working copy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 vendor
 composer.lock
+.idea


aik099:
I have it in global `.gitignore`, but having it in repo-specific `.gitignore` won't hurt.

jpastoor:
Did not know that was possible, nice tip.

PR #72, opened 2016-02-28
Improved docblocks and method (parameter) names in the API class

jpastoor:
Not complete in anyway, but I think a step in the right directory.
- Renamed setEndPoint to setEndpoint to be in line with the getter and naming convention. (Not BC since method names are case insensitive)
- Added types to most docblock parameters and return values
- Removed implicit variable assignments in getCreateMeta for improved clarity and IDE support
- Removed all PHPStorm warnings. 
- Changed some textual stuff

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -36,7 +36,6 @@ class Api
     const REQUEST_PUT = 'PUT';
     const REQUEST_DELETE = 'DELETE';
 
-
     const AUTOMAP_FIELDS = 0x01;
 
     /** @var string $endpoint */
@@ -61,9 +60,9 @@ class Api
     protected $statuses;
 
     /**
-     * create a jira api client.
+     * Create a JIRA API client.
      *
-     * @param $endpoint
+     * @param string $endpoint URL
      * @param AuthenticationInterface $authentication
      * @param ClientInterface $client
      */
@@ -82,15 +81,18 @@ class Api
         $this->client = $client;
     }
 
+    /**
+     * @param int $options
+     */
     public function setOptions($options)
     {
         $this->options = $options;
     }
 
     /**
-     * get endpoint url
+     * Get Endpoint URL
      *
-     * @return mixed
+     * @return string
      */
     public function getEndpoint()
     {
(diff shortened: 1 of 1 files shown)


src/Jira/Api.php
```diff
@@ -61,9 +60,9 @@ class Api
     protected $statuses;
 
     /**
-     * create a jira api client.
+     * Create a JIRA API client.
      *
-     * @param $endpoint
+     * @param string $endpoint URL
      * @param AuthenticationInterface $authentication
```
aik099: Any purpose in not adding parameter description to other method parameters?

Same is true for other DocBlocks.

src/Jira/Api.php
```diff
@@ -82,15 +81,18 @@ public function __construct(
         $this->client = $client;
     }
 
+    /**
+     * @param int $options
```
aik099: Not sure if short description in mandatory in DocBlock but I usually always add it.

Same is true for other DocBlocks.

src/Jira/Api.php
```diff
@@ -131,54 +133,75 @@ public function getFields()
...
+     * @param string $projectKey
+     *
+     * @return array|false
+     */
     public function getProject($projectKey)
     {
-        $result = $this->api(self::REQUEST_GET, "/rest/api/2/project/{$projectKey}", array(), true);
-
-        return $result;
+        return $this->api(self::REQUEST_GET, "/rest/api/2/project/{$projectKey}", array(), true);
```
aik099: The `$projectKey` variable can be extracted from double-quoted string to allow single quote usage as you did above in `getAttachment` method.

aik099:
Today’s review completed.

src/Jira/Api.php
jpastoor: As said in the PR its a step in the right direction I think, but in no way complete. Did it in the constructor of API since that variable might be vague for some new to the lib.

aik099: OK.

jpastoor:
Thanks & I agree for your comments. I'd suggest merging this one and I'll have a go at further documenting other classes and stuff you commented on in a followup PR.

aik099:
OK. Merging, thank you @jpastoor .

PR #76, opened 2016-03-28
Support for DELETE requests in the "CurlClient" class

zimmermanj42:
The current CURL implementation does not acknowledge DELETE requests.  This fix allows for the HTTP DELETE verb to be used with the CURL wrapper.

However, all DELETE requests return "(bool) false" from within the "Api::api()" function call.  It appears that the Api class was not built really with DELETE calls in mind.  If the "Api::api()" function specifies how DELETE requests should return, I could also probably fix that as well.

### DIFF
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -94,8 +94,8 @@ class CurlClient implements ClientInterface
             } else {
                 curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($data));
             }
-        } elseif ($method == 'PUT') {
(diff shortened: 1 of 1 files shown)


aik099:
Unless the DELETE requests have any POST parameters please change implementation to one from here: https://github.com/chobie/jira-api-restclient/pull/53/files#diff-bac9579c3d575193a096385b414bc909 (this is diff from not yet merged PR).

aik099:

aik099:
Your changes were incorporated into #108. Thank you, @zimmermanj42 .

aik099:
Changes moved to #115.

PR #77, opened 2016-03-28
Allows for "expand" option in search requests.

zimmermanj42:
Since this utilizes NULL for the default, it should not create any BC breaks. Also, the underlying HTTP transmission code utilizes "http_build_query()" from PHP, and in this function array entries with NULL values are not included in the output. So, this change wouldn't even affect the URLs that are generated.

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -440,10 +440,11 @@ class Api
      * @param int $startAt
      * @param int $maxResult
      * @param string $fields
+     * @param string $expand
      *
      * @return Result|false
      */
-    public function search($jql, $startAt = 0, $maxResult = 20, $fields = '*navigable')
+    public function search($jql, $startAt = 0, $maxResult = 20, $fields = '*navigable', $expand = null)
     {
         $result = $this->api(
             self::REQUEST_GET,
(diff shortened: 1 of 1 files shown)


zimmermanj42:
I created this PR as I was trying to get Jira issue descriptions to output as HTML.  I first tried using a Markdown parser, but it looks like Jira uses it's own proprietary markup language.  As such, the only way to get the HTML version of the description would be to allow for the "expand" field to be added to the API request.

For information, read the "Expansion" section on the API documentation: https://docs.atlassian.com/jira/REST/latest/

src/Jira/Api.php
```diff
@@ -453,6 +454,7 @@ public function search($jql, $startAt = 0, $maxResult = 20, $fields = '*navigabl
                 'startAt' => $startAt,
                 'maxResults' => $maxResult,
                 'fields' => $fields,
+                'expand' => $expand,
```
aik099: Better not to rely on fact, that `http_build_query` is used in used in underlying client class and do the following:

1. move out the parameter array into a separate variable
2. add an IF statement that would check for `isset($expand)` and only then add array entry
3. use that array in `->api` call

aik099:

Issue #74, opened 2016-03-02
Create integration tests


aik099:
For integration tests need to create class, that would:

* implement `ClientInterface`
* return stubbed responses based on API docs https://docs.atlassian.com/jira/REST/latest/

I have no idea right now how to test actual `ClientInterface` implementations though.

Actually we might create mini JIRA REST API endpoint that would return stubbed responses instead. Then we can run all tests through each client (CurlClient and PHPClient) to see if they actually works.

aik099:
Decided to take different route. In #108 we have integration tests for `ClientInterface` implementing classes. Then in #114, since we know that client classes work as expected, the mock for `ClientInterface` is created and we test `Api` class.

PR #82, opened 2016-05-02
Added updateVersion, releaseVersion and findVersionByName

jpastoor:
Expanded a bit on the API's functionality for dealing with Project Versions.
Methods are unittested and work for the current live version of JIRA hosted.

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -26,8 +26,8 @@ namespace chobie\Jira;
 
 use chobie\Jira\Api\Authentication\AuthenticationInterface;
 use chobie\Jira\Api\Client\ClientInterface;
-use chobie\Jira\Api\Result;
 use chobie\Jira\Api\Client\CurlClient;
+use chobie\Jira\Api\Result;
 
 class Api
 {
@@ -222,6 +222,7 @@ class Api
      * @param $issuetypeIds array    Combined with issuetypeNames, lists the issue types with which to filter the results.
      *                               If null, all issue types are returned. Specifiying an issue type that does not exist is
      *                               not an error.
+     *
      * @param $issuetypeNames array  Combined with issuetypeIds, lists the issue types with which to filter the results.
      *                               If null, all issue types are returned. This parameter can be specified multiple times,
      *                               but is NOT interpreted as a comma-separated list. Specifiying an issue type that does
@@ -344,6 +345,33 @@ class Api
     public function getVersions($projectKey)
     {
         return $this->api(self::REQUEST_GET, "/rest/api/2/project/{$projectKey}/versions", array(), true);
+    }
+
+    /**
+     * Helper method to find a specific version based on the name of the version.
+     *
+     * @param string $projectKey Project Key
+     * @param string $name The version name to match on
+     *
+     * @return int|null VersionId on match or null when there is no match
+     */
+    public function findVersionByName($projectKey, $name)
+    {
+        // Fetch all versions of this project
+        $versions = $this->getVersions($projectKey);
+
+        // Filter results on the name
+        $matching_versions = array_filter($versions, function (array $version) use ($name) {
+            return $version['name'] == $name;
+        });
+
+        // Early out for no results
+        if (empty($matching_versions)) {
+            return null;
+        }
+
+        // Multiple results should not happen since name is unique
+        return reset($matching_versions);
     }
 
     /**
@@ -483,6 +511,47 @@ class Api
         );
 
         return $this->api(self::REQUEST_POST, '/rest/api/2/version', $options);
+    }
+
+    /**
+     * Update JIRA Version
+     *
+     * https://docs.atlassian.com/jira/REST/latest/#api/2/version-updateVersion
+     * 
+     * @param int $versionId Version identifier
+     * @param array $params Key->Value list to update the version with. 
+     *
+     * @return Result|false
+     */
+    public function updateVersion($versionId, $params = array())
+    {
+        return $this->api(self::REQUEST_PUT, sprintf('/rest/api/2/version/%d', $versionId), $params);
+    }
+
+    /**
+     * Shorthand to mark a version as Released
+     *
+     * @param int $versionId  Version identifier
+     * @param string|null $releaseDate Date in Y-m-d format. Defaults to today
+     * @param array $params Optionally extra parameters.
+     *
+     * @return Result|false
+     */
+    public function releaseVersion($versionId, $releaseDate = null, $params = array())
+    {
+        if(!$releaseDate) {
+            $releaseDate = date('Y-m-d');
+        }
+
+        $params = array_merge(
+            array(
+                'releaseDate' => $releaseDate,
+                'released' => true
+            ),
+            $params
+        );
+
+        return $this->updateVersion($versionId, $params);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


aik099:
Please fix the build. I guess you're using PHP 5.5 syntax.

src/Jira/Api.php
```diff
@@ -486,6 +514,45 @@ public function createVersion($projectId, $name, $options = array())
...
+     * Update JIRA Version
+     *
+     * https://docs.atlassian.com/jira/REST/latest/#api/2/version-updateVersion
+     * 
+     * @param int $versionId Version identifier
+     * @param array $params Key->Value list to update the version with. 
+     *
+     * @return Result|false
+     */
+    public function updateVersion($versionId, $params = array()) {
```
aik099: Please be consistent about `{` usage in method declaration. Other methods place `{` on next line.

I recommend forcing IDE to check for such stuff.

src/Jira/Api.php
```diff
@@ -486,6 +514,45 @@ public function createVersion($projectId, $name, $options = array())
...
+     *
+     * @param int $versionId  Version identifier
+     * @param string|null $releaseDate Date in Y-m-d format. Defaults to today
+     * @param array $params Optionally extra parameters.
+     *
+     * @return Result|false
+     */
+    public function releaseVersion($versionId, $releaseDate = null, $params = array()) {
+        if(!$releaseDate) {
+            $releaseDate = date("Y-m-d");
```
aik099: Double quotes are not needed in there.

src/Jira/Api.php
```diff
@@ -486,6 +514,45 @@ public function createVersion($projectId, $name, $options = array())
...
+        }
+
+        $params = array_merge(
+            array(
+                "releaseDate" => $releaseDate,
```
aik099: Double quotes are not needed in here.

tests/Jira/ApiTest.php
```diff
@@ -25,4 +25,102 @@ public function testSetEndpointTrailingSlash()
...
+    public function testUpdateVersion()
+    {
+        $params = array(
+            'released' => true,
+            'releaseDate' => '2010-07-06',
+        );
+
+        // Stub the api method and keep the rest intact
+        /** @var Api|\PHPUnit_Framework_MockObject_MockObject $api */
+        $api = $this->getMockBuilder('\chobie\Jira\Api')->setMethods(array('api'))->disableOriginalConstructor()->getMock();
```
aik099: We can use Prophecy now, so why not use it for mock creation.

aik099: This test doesn't really test anything except out internal cross-class interactions 😄 

You can keep it of course, but ideally I would like to see #74 implemented at some point.

src/Jira/Api.php
```diff
@@ -347,6 +348,33 @@ public function getVersions($projectKey)
...
+     * @param string $name The version name to match on
+     *
+     * @return int|null VersionId on match or null when there is no match
+     */
+    public function findVersionByName($projectKey, $name)
+    {
+        // Fetch all versions of this project
+        $versions = $this->getVersions($projectKey);
+
+        // Filter on name in the PHP (Can't be done using the API?)
```
aik099: Nope, there is no way. You can drop this comment.

tests/Jira/ApiTest.php
jpastoor: Not sure why that is better than using the native phpunit way?

src/Jira/Api.php
jpastoor: Fixed

src/Jira/Api.php
jpastoor: Fixed

src/Jira/Api.php
jpastoor: Fixed

src/Jira/Api.php
```diff
@@ -347,6 +348,33 @@ public function getVersions($projectKey)
...
 
     /**
+     * Helper method to find a specific version based on the name of the version.
+     *
+     * @param string $projectKey Project Key
+     * @param string $name The version name to match on
+     *
+     * @return int|null VersionId on match or null when there is no match
+     */
+    public function findVersionByName($projectKey, $name)
```
aik099: I recommend setting `private` scope for new methods, that you don't plan to expose in the API yet.

Adding all stuff public by default work create more problems down the road, when we release 2.0 version and need to maintain that API to avoid BC breaks.

src/Jira/Api.php
```diff
@@ -486,6 +514,47 @@ public function createVersion($projectId, $name, $options = array())
...
 
     /**
+     * Update JIRA Version
+     *
+     * https://docs.atlassian.com/jira/REST/latest/#api/2/version-updateVersion
+     * 
+     * @param int $versionId Version identifier
+     * @param array $params Key->Value list to update the version with. 
+     *
+     * @return Result|false
```
aik099: Are you sure, that `Result` object is ever returned on PUT requests?

jpastoor:
Fixed style issues, will do a rebase before landing.

tests/Jira/ApiTest.php
aik099: You know that PHPUnitMocks isn't that powerful and has it's own limitations. Since PHPUnit 4.8 the built-in support for Prophecy was added and people are encouraged to use it for mock creation.

It also encourages good architecture but at least not allowing to partially mock objects (e.g. `api` method call gets mocked, while others are not as you did above).

jpastoor: Agreed.

src/Jira/Api.php
aik099: In this method yes, but you've added other methods as well and they have same problem.

src/Jira/Api.php
jpastoor: This method is meant to be public. (and is not used internally in the lib yet) 
I'm using it in a project to search the corresponding release based on a git tag :)

aik099: Ah, then maybe other 2 should not be.

src/Jira/Api.php
jpastoor: Yep, a struct like

class chobie\Jira\Api\Result#5 (5) {
  protected $expand =>
  NULL
  protected $startAt =>
  NULL
  protected $maxResults =>
  NULL
  protected $total =>
  NULL
  protected $result =>
  array(8) {
    'self' =>
    string(56) "https://xxx.atlassian.net/rest/api/2/version/14206"
    'id' =>
    string(5) "14206"
    'name' =>
    string(6) "3.63.0"
    'archived' =>
    bool(false)
    'released' =>
    bool(true)
    'releaseDate' =>
    string(10) "2016-05-02"
    'userReleaseDate' =>
    string(9) "02/May/16"
    'projectId' =>
    int(10100)
  }
}

aik099: OK.

src/Jira/Api.php
jpastoor: Fixed

src/Jira/Api.php
jpastoor: I think we should add these methods since they are pretty basic functionality for anyone who wants to automate the release process using using JIRA Versions/Releases.

aik099: OK. But in either case until 2.0.0 release all methods not present in 1.x release are subject to removal if such need arises.

PR #83, opened 2016-05-05
JIRA response will have 'errors' as key

mmkader85:
https://docs.atlassian.com/jira/REST/latest/#api/2/filter-createFilter

### DIFF
--- a/src/Jira/Api/Result.php
+++ b/src/Jira/Api/Result.php
@@ -68,9 +68,9 @@ class Result
 
     public function getIssues()
     {
-        if (isset($this->result['issues'])) {
+        if (isset($this->result['errors'])) {
             $result = array();
-            foreach ($this->result['issues'] as $issue) {
+            foreach ($this->result['errors'] as $issue) {
                 $result[] = new Issue($issue);
             }
             return $result;


aik099:
So you're saying that actual issues are returned in `errors` key of REST API response?

aik099:
Maybe the request, that's response you're feeding into `Result` class has something wrong in it?

mmkader85:
Yes, if there are any errors then it is returing with the key 'errors' and not 'issues'. See a sample response from JIRA below.


```php
chobie\Jira\Api\Result Object
(
    [expand:protected] =>
    [startAt:protected] =>
    [maxResults:protected] =>
    [total:protected] =>
    [result:protected] => Array
        (
            [errorMessages] => Array
                (
                )
            [errors] => Array
                (
                    [customfield_10190] => Option value 'Sigapore' is not valid
                    [customfield_10301] => Option value 'We' is not valid
                )
        )
)
```

aik099:
But method name you've changed is called `getIssues` it it returns issues. When there are errors, then no issues are returned and that's absolutely correct behavior.

If you want to get errors you can do 2 things:

1. either use `getResult` and manually look for errors
2. add `getErrors` method that would return errors if any and empty array otherwise

aik099:
I hope you do understand that proposed change would result in `getIssues` method returning an  empty array instead of found issues, when no errors happened 😄

mmkader85:
Ah, yes. Realised it. Confused JIRA issues(Ticket) with errors. Thanks mate.

PR #85, opened 2016-05-09
Added getErrors()

mmkader85:
When JIRA response contains errors, this can be used.

### DIFF
--- a/src/Jira/Api/Result.php
+++ b/src/Jira/Api/Result.php
@@ -83,11 +83,13 @@ class Result
     {
     	if (isset($this->result['errors'])) {
             $result = array();
-            foreach ($this->result['errors'] as $issue) {
-                $result[] = new Issue($issue);
+            foreach ($this->result['errors'] as $error) {
+                $result[] = $error;
             }
             return $result;
         }
+        
+        return array();
     }
 
     public function getResult()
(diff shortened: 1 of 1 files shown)


src/Jira/Api/Result.php
```diff
@@ -78,6 +78,17 @@ public function getIssues()
 
         return array();
     }
+    
+    public function getErrors()
+    {
+    	if (isset($this->result['errors'])) {
+            $result = array();
+            foreach ($this->result['errors'] as $issue) {
+                $result[] = new Issue($issue);
```
aik099: Why you need to create `Issue` object with error content? Each entry in `errors` array doesn't represent an issue at all.

src/Jira/Api/Result.php
```diff
@@ -78,6 +78,17 @@ public function getIssues()
 
         return array();
     }
+    
+    public function getErrors()
+    {
+    	if (isset($this->result['errors'])) {
```
aik099: What would make sense instead is to just return value of `$this->result['errors']` as is, when it's set.

aik099:

aik099:
Sorry, but after pondering a bit on this I started to believe, that this new method won't be of any help.

PR #86, opened 2016-05-19
Fixing deprecated error when uploading file attachments in PHP 5.5+

DerMika:
Same pull request as done on jpastoor/jira-api-restclient:

On PHP 5.5 +, you get an `E_DEPRECATED` when trying to upload files. This PR fixes that by using a `\CURLFile` object to upload (see https://wiki.php.net/rfc/curl-file-upload)

### DIFF
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -90,6 +90,7 @@ class CurlClient implements ClientInterface
         if ($method == 'POST') {
             curl_setopt($curl, CURLOPT_POST, 1);
             if ($isFile) {
+                $data['file'] = $this->getCurlValue($data['file']);
                 curl_setopt($curl, CURLOPT_POSTFIELDS, $data);
             } else {
                 curl_setopt($curl, CURLOPT_POSTFIELDS, json_encode($data));
@@ -121,5 +122,20 @@ class CurlClient implements ClientInterface
         }
 
         return $data;
+    }
+	
+    /**
+      * If necessary, replace curl file @ string with a CURLFile object (for PHP 5.5 and up)
+      *
+      * @param string $fileString The string in @-format as it is used on PHP 5.4 and older.
+      * @return \CURLFile|string
+      */
+    protected function getCurlValue($fileString)
+    {
+        if (!function_exists('curl_file_create')) {
+            return $fileString;
+        }
+
+        return curl_file_create(substr($fileString, 1));
     }
 }


src/Jira/Api/Client/CurlClient.php
```diff
@@ -90,6 +90,7 @@ public function sendRequest(
         if ($method == 'POST') {
             curl_setopt($curl, CURLOPT_POST, 1);
             if ($isFile) {
+				$data['file'] = $this->getCurlValue($data['file']);
```
aik099: Wrong indentation.

DerMika: Woops, I'll fix it

src/Jira/Api/Client/CurlClient.php
```diff
@@ -122,4 +123,27 @@ public function sendRequest(
 
         return $data;
     }
+	
+	/**
```
aik099: Wrong indentation.

src/Jira/Api/Client/CurlClient.php
```diff
@@ -122,4 +123,27 @@ public function sendRequest(
...
+      * @param string $fileString The string in @-format as it is used on PHP 5.4 and older.
+      * @return \CURLFile|string
+      */
+     protected function getCurlValue($fileString)
+     {
+         $theCurlValue = $fileString;
+         $regex = "|^@(.*);filename=(.*)$|";
+         $matches = array();
+ 
+         if (function_exists('curl_file_create') && preg_match($regex, $fileString, $matches) === 1) {
```
aik099: When `curl_file_create` function doesn't exist exit immediately at the method beginning.

src/Jira/Api/Client/CurlClient.php
```diff
@@ -122,4 +123,27 @@ public function sendRequest(
...
+	/**
+      * If necessary, replace curl file @ string with a CURLFile object (for PHP 5.5 and up)
+      *
+      * @param string $fileString The string in @-format as it is used on PHP 5.4 and older.
+      * @return \CURLFile|string
+      */
+     protected function getCurlValue($fileString)
+     {
+         $theCurlValue = $fileString;
+         $regex = "|^@(.*);filename=(.*)$|";
```
aik099: Is this regex for checking if given value isn't already in correct format that is created by `curl_file_create` function?

src/Jira/Api/Client/CurlClient.php
```diff
@@ -122,4 +123,27 @@ public function sendRequest(
...
+         $matches = array();
+ 
+         if (function_exists('curl_file_create') && preg_match($regex, $fileString, $matches) === 1) {
+             $fileName = $matches[1];
+             $postName = $matches[2];
+             $contentType = null; //Not used in \chobie\Jira\Api at the moment
+ 
+             $theCurlValue = curl_file_create($fileName, $contentType, $postName);
```
aik099: Just return it right here.

src/Jira/Api/Client/CurlClient.php
aik099: If not preg_matched:

1. return original value 
2. add comment explaining why this happened (e.g. value is already properly encoded)

src/Jira/Api/Client/CurlClient.php
```diff
@@ -122,4 +123,27 @@ public function sendRequest(
...
+      * @return \CURLFile|string
+      */
+     protected function getCurlValue($fileString)
+     {
+         $theCurlValue = $fileString;
+         $regex = "|^@(.*);filename=(.*)$|";
+         $matches = array();
+ 
+         if (function_exists('curl_file_create') && preg_match($regex, $fileString, $matches) === 1) {
+             $fileName = $matches[1];
```
aik099: I'm confused. Does anybody even pre-encode value currently? Maybe all that pregmatching isn't even needed?

DerMika: I believe I've made a mistake: In my original implementation I added an extra parameter to `\chobie\Jira\Api::createAttachment` to provide a name for the file to be uploaded. I did that in the format `'@' . $filename . ';filename=' . $postname` and I simply copy-pasted the function from my own implementation. I'll remove the clutter.

DerMika:
Ok, removed all the unneeded stuff. Is this more like it?

src/Jira/Api/Client/CurlClient.php
```diff
@@ -122,4 +123,19 @@ public function sendRequest(
...
+      *
+      * @param string $fileString The string in @-format as it is used on PHP 5.4 and older.
+      * @return \CURLFile|string
+      */
+    protected function getCurlValue($fileString)
+    {
+        if (!function_exists('curl_file_create')) {
+            return $fileString;
+        } else {
+            return curl_file_create(substr($fileString, 1));
```
aik099: Just return it, no `else` needed.

aik099:
> Ok, removed all the unneeded stuff. Is this more like it?

Almost.

aik099:
In which PHP version `@filename` notation would stop working for curl uploads?

DerMika:
Here is the RFC: https://wiki.php.net/rfc/curl-file-upload

they say 

> In 5.6, @ option will be switched off by default, but can still be enabled by explicit curl_setopt setting, such as:

and 

> In future versions, this capability may be removed completely.

But I don't think an actual decision is made to remove this capability yet.

aik099:
So without your change on PHP 5.6 file upload won't work at all, right?

DerMika:
On PHP 5.6 you can still enable the old behaviour with a `curl_setopt(CURLOPT_SAFE_UPLOAD, false);` but even then you get the `E_DEPRECATED` - which is the one that was bugging me.

aik099:
> On PHP 5.6 you can still enable the old behaviour with a curl_setopt(CURLOPT_SAFE_UPLOAD, false); but even then you get the E_DEPRECATED - which is the one that was bugging me.

You can enable it, but it's turned off by default and it means that uploads through this JIRA API client won't work anymore unless I merge your fix.

DerMika:
Well actually it's currently already enabled by default - see https://github.com/chobie/jira-api-restclient/blob/master/src/Jira/Api/Client/CurlClient.php#L83 : 

```
        if ($isFile) {
            if (defined('CURLOPT_SAFE_UPLOAD')) {
                curl_setopt($curl, CURLOPT_SAFE_UPLOAD, false);
            }
```

As I said, my PR was just to get rid of the ugly deprecated notice. And also to be future proof once this option goes away in the future.

aik099:
Then maybe setting `CURLOPT_SAFE_UPLOAD` to `false` isn't needed anymore?

DerMika:
I would think so, but honestly I haven't investigated if that could cause adverse effects.

aik099:
OK. Then could you please squash and I'll merge it in.

DerMika:
Unless I did something wrong, I believe I've squashed everything already?

aik099:
Merging. Thanks @DerMika .

DerMika:
You're welcome, thanks for indulging me :)

PR #87, opened 2016-05-20
Fix uploaded attachment's filename

DerMika:
With the current implementation, an attachment you upload to JIRA gets the name of the full path you specified for the file upload location. 

In the example https://github.com/chobie/jira-api-restclient/blob/master/examples/create_attachement.php this problem cannot be seen because the sample uses a file to upload from the same folder using a relative path, but if you start using absolute paths, your attachments in JIRA will have very ugly and potentially long filenames - not to mention that it may expose file system information you probably don't want to leak.

This PR fixes that by setting the `; filename=` (pre-PHP 5.5) or the `postname` parameter of `\CURLFile` (PHP 5.5+) to just the file without any path info.

### DIFF
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -133,9 +133,9 @@ class CurlClient implements ClientInterface
     protected function getCurlValue($fileString)
     {
         if (!function_exists('curl_file_create')) {
-            return $fileString;
+            return $fileString . '; filename=' . basename($fileString);
         }
 
-        return curl_file_create(substr($fileString, 1));
+        return curl_file_create(substr($fileString, 1), null, basename($fileString));
     }
 }


aik099:
Was is working correctly before change you've introduced in #86?

DerMika:
No, the same issue existed.

DerMika:
It's a general issue with CURL uploads - see for example http://stackoverflow.com/questions/15127732/change-name-of-upload-file-in-curl

aik099:
I wonder why nobody complained about this fact before. Same issue with PhpClient?

DerMika:
No in fact, the PHPClient does the same thing as what I'm doing in this PR: https://github.com/chobie/jira-api-restclient/blob/master/src/Jira/Api/Client/PHPClient.php#L101

`'Content-Disposition: form-data; name="file"; filename="' . basename($filename) . "\"\r\n" .`

aik099:
I see now. Since this issue wasn't reported before people are:

* either not uploading much of the files
* or using `PHPClient`

aik099:
Merging. Thanks again @DerMika .

PR #100, opened 2016-06-08
Removed incorrect typehint of $data in Api::api() which caused setWatchers to throw a warning

alopex06:
setWatchers Method pass a String as Data for the api call and not an array. This produces an warning that crash some applications.

PHP Error:

PHP Catchable fatal error:  Argument 3 passed to chobie\Jira\Api::api() must be an array, string given

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -691,7 +691,7 @@ class Api
 	public function api(
 		$method = self::REQUEST_GET,
 		$url,
-		array $data = array(),
+		$data = array(),
 		$return_as_json = false,
 		$is_file = false,
 		$debug = false
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -681,7 +681,7 @@ class Api
 	 *
 	 * @param string  $method         Request method.
 	 * @param string  $url            URL.
-	 * @param array   $data           Data.
+	 * @param array|string   $data           Data.
 	 * @param boolean $return_as_json Return results as JSON.
 	 * @param boolean $is_file        Is file-related request.
 	 * @param boolean $debug          Debug this request.
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -679,12 +679,12 @@ class Api
 	/**
 	 * Send request to specified host.
 	 *
-	 * @param string  $method         Request method.
-	 * @param string  $url            URL.
-	 * @param array|string   $data           Data.
-	 * @param boolean $return_as_json Return results as JSON.
-	 * @param boolean $is_file        Is file-related request.
-	 * @param boolean $debug          Debug this request.
+	 * @param string       $method         Request method.
+	 * @param string       $url            URL.
+	 * @param array|string $data           Data.
+	 * @param boolean      $return_as_json Return results as JSON.
+	 * @param boolean      $is_file        Is file-related request.
+	 * @param boolean      $debug          Debug this request.
 	 *
 	 * @return array|Result|false
 	 */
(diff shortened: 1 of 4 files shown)


aik099:
Interesting. More issues related to this, that I've found:

1. set `setWatchers` method is the only one, that does so
2. the `ClientInterface::sendRequest` method (in all implementations) that is called from `Api::api` when we're doing GET request is expecting `$data` to be array (passes it into `http_build_query`), but don't care in all other request methods (e.g. POST)

jpastoor:
https://docs.atlassian.com/jira/REST/latest/#api/2/issue-addWatcher
seems indeed valid to add as string, so I don't think we can enforce data to be an array. (regression in 49615cf)

aik099:
The change (when typehint was added) was a result of wrong DocBlock comment on the `sendRequest` and `api` methods. To avoid such mistake in future I'm proposing:

1. update data type (+ reformat DocBlock) for `Api::api`
2. update data type (+ reformat DocBlock) for all `sendRequest` methods (2 implementation + interface)
3. in `sendRequest` implementations throw an `InvalidArgumentException` when GET request is made, but `$data` isn't an array

jpastoor:
Agree with aik099 points and would very much like to see that in this PR.

Some additional wishes (but if you aint sure how to do it I'll add it soon in a separate PR)
4. Unittest to validate 3.
5. Unittest to validate setWatcher propagates correctly to the sendRequest interface

src/Jira/Api.php
```diff
@@ -681,7 +681,7 @@ public function createRemotelink(
 	 *
 	 * @param string  $method         Request method.
 	 * @param string  $url            URL.
-	 * @param array   $data           Data.
+	 * @param array|string   $data           Data.
```
aik099: Please reformat DocBlock to alight parameter names and their descriptions.

src/Jira/Api/Client/CurlClient.php
```diff
@@ -75,6 +75,10 @@ public function sendRequest(
 
 		if ( $method == 'GET' ) {
 			$url .= '?' . http_build_query($data);
+
+			if ( !is_array($data) ) {
+				throw new \InvalidArgumentException('Data must be an array.');
```
aik099: Please mention in DocBlock via `@throws` tag, that this exception is being thrown.

aik099:
Also `PHPClient` and `ClientInterface` where not changed.

alopex06:
Ok thank you for the quick responses.
@jpastoor that would be nice if you could update the unittests later.

aik099:
> @jpastoor that would be nice if you could update the unittests later.

Not used to writing unit tests?

jpastoor:
Merged, thanks

src/Jira/Api/Client/CurlClient.php
```diff
@@ -75,6 +76,10 @@ public function sendRequest(
 
 		if ( $method == 'GET' ) {
 			$url .= '?' . http_build_query($data);
+
+			if ( !is_array($data) ) {
```
aik099: This needs to be moved up. Otherwise you already get notice about wrong parameter usage in `http_build_query` call.

aik099:
@jpastoor , this PR is incomplete. It not only misses tests, but also misses points we've mentioned as needed.

Therefore until you check that all requested items were done don't merge it.

src/Jira/Api/Client/CurlClient.php
aik099: And same part in `PHPClient` is still missing.

PR #101, opened 2016-06-15
Add call for getting attachments meta configuration

N-M:
Option to get attachments meta configuration.

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -202,7 +202,6 @@ class Api
 	 *
 	 * @return array
 	 */
-	public function getAttachmentUploadLimit()
 	{
 		return $this->api(self::REQUEST_GET, '/rest/api/2/attachment/meta', array());
 	}
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -202,6 +202,7 @@ class Api
 	 *
 	 * @return array
 	 */
+	public function getAttachmentsMeta()
 	{
 		return $this->api(self::REQUEST_GET, '/rest/api/2/attachment/meta', array());
 	}
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -198,6 +198,16 @@ class Api
 	}
 
 	/**
+	 * Gets attachment upload limit.
+	 *
+	 * @return array
+	 */
+	public function getAttachmentUploadLimit()
+	{
+		return $this->api(self::REQUEST_GET, '/rest/api/2/attachment/meta', array());
+	}
+
+	/**
 	 * Gets attachment.
 	 *
 	 * @param string $attachment_id Attachment ID.
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -198,11 +198,12 @@ class Api
 	}
 
 	/**
-	 * Gets attachment upload limit.
+	 * Gets attachments meta information.
 	 *
 	 * @return array
+	 * @since  2.0.0
 	 */
-	public function getAttachmentsMeta()
+	public function getAttachmentsMetaInformation()
 	{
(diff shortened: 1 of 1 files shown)


src/Jira/Api.php
```diff
@@ -198,6 +198,16 @@ public function editIssue($issue_key, array $params)
 	}
 
 	/**
+	 * Gets attachment upload limit.
+	 *
+	 * @return array
+	 */
+	public function getAttachmentUploadLimit()
```
aik099: Please rename into `getAttachmentsMeta`, because according to https://docs.atlassian.com/jira/REST/latest/#api/2/attachment-getAttachmentMeta not only upload limit is returned but all attachment configuration set in given JIRA instance.

src/Jira/Api.php
```diff
@@ -202,7 +202,6 @@ public function editIssue($issue_key, array $params)
 	 *
 	 * @return array
 	 */
-	public function getAttachmentUploadLimit()
```
aik099: 😄  I said renaming method. Not removing 1st line of it's declaration.

src/Jira/Api.php
```diff
@@ -198,6 +198,15 @@ public function editIssue($issue_key, array $params)
 	}
 
 	/**
+	 * Gets attachment upload limit.
```
aik099: Also the short description needs to be updated to reflect method name change.

src/Jira/Api.php
```diff
@@ -198,6 +198,16 @@ public function editIssue($issue_key, array $params)
...
+	 *
+	 * @return array
+	 */
+	public function getAttachmentsMeta()
```
aik099: Please respect both your and my time. If I say name method like that, then please do name it exactly like it or state your objections, why this can't be done.

N-M: Sorry for confusion is this fine?
`/**
	 * Gets attachment meta information.
	 *
	 * @return array
	 * @since  2.0.0
	 */
	public function getAttachmentsMetaInformation()`

aik099: Almost. Just change `Gets attachment meta information.` into `Gets attachments meta information.` (added `s`).

src/Jira/Api.php
```diff
@@ -198,6 +198,17 @@ public function editIssue($issue_key, array $params)
...
 
 	/**
+	 * Gets attachments meta information.
+	 *
+	 * @return array
+	 * @since  2.0.0
+	 */
+	public function getAttachmentsMetaInformation()
+	{
+		return $this->api(self::REQUEST_GET, '/rest/api/2/attachment/meta', array());
```
aik099: Isn't the 3rd argument default value also `array()`? If that is so, then you can just omit it.

N-M: You are correct it is removed.

aik099:
It's kind of stupid to ask at this point, but is knowing upload limit before making API calls really helps?

Otherwise please squash and it's ready to be merged.

N-M:
In my use case i need to present a user upload limit before he chooses to add ticket with attachments on Jira.

N-M:
In my use case i need to present a user upload limit before he chooses to add ticket with attachments on Jira.

I do not see squash option how i can do it?

aik099:
> I do not see squash option how i can do it?

Where do you want to see it? This is a way to combine all commits on a branch (that powers the PR) onto single commit. This is pure git feature.

In general case:

1. run this command: `git rebase -i HEAD~3`  (the 3 is number of commits you want to combine)
2. in dialog that opens select squash option for all but 1st commit
3. change commit message to be used for final commit on next screen

Actually GitHub now has this feature as well, when merging PR. Let me try that.

aik099:
Merged, thanks @N-M .

PR #105, opened 2016-06-24
Refactoring for PSR-2 standars

ptondereau:
Hello,

I've refactored src and tests to be more psr2 compliant.
I've changed namespaces so It's a BC.

I've added PHP7 to travis and activate the composer cache for boosting travis results.
Finally, I've added prefer-stable to true in composer.json because I think, you should only use stable versions for dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,25 @@
 language: php
+
 php:
   - 5.3
   - 5.4
   - 5.5
+  - 5.6
+  - 7.0
+  - hhvm
 
-install:
-  - composer install --no-interaction
+# This triggers builds to run on the new TravisCI infrastructure.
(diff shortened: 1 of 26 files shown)


aik099:
Thank you for your time, but next time please follow contrtibution instructions from https://github.com/chobie/jira-api-restclient/blob/master/CONTRIBUTING.md page.

This PR won't be merged, because it:

* changes coding standard in this project to different one, even when library already using another standard
* does multiple unrelated things in one PR (not an on topic PR), like changing Travis CI config and changing standard

To make use of this PR please do the following:

1. create an issue about every thing you think needs changing and explain in there why this needs to be changed
2. once confirmed by library maintainers you can proceed with PR sending against each of the issues created above

aik099:
Changes related to adding new PHP versions and enabling Travis CI caches were implemented as separate PRs.

PR #114, opened 2016-07-01
Use Prophecy for mocking in "Api" class tests

aik099:
Problems solved:

* the DocBlock of `updateVersion` and `releaseVersion` methods were saying that they return `Result`, but they never do
* the DocBlock of `findVersionByName` method was saying that it returns VersionId, but it actually returns all version fields
* tests for `Api` class were mocking `Api` class partially and that is considered as bad practice (changed to mock client class instead)
* several assertions per test were replaced with data provider usage (where appropriate)
* removed DocBlocks of test methods in favor of improved test names themselves
* added helper methods to be used for quick test writing for all other methods of `Api` class

P.S.
My apologies for 3 places, where PhpStorm removed trailing whitespaces. Not sure if it's worth redoing PR to exclude them.

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -426,7 +426,7 @@ class Api
 	 * @param string $project_key Project Key.
 	 * @param string $name        The version name to match on.
 	 *
-	 * @return integer|null VersionId on match or null when there is no match.
+	 * @return array|null Version data on match or null when there is no match.
 	 * @since  2.0.0
 	 */
 	public function findVersionByName($project_key, $name)
@@ -592,7 +592,7 @@ class Api
 	 * @param integer $version_id Version ID.
 	 * @param array   $params     Key->Value list to update the version with.
 	 *
-	 * @return Result|false
+	 * @return false
 	 * @since  2.0.0
 	 * @link   https://docs.atlassian.com/jira/REST/latest/#api/2/version-updateVersion
 	 */
@@ -608,7 +608,7 @@ class Api
 	 * @param string|null $release_date Date in Y-m-d format (defaults to today).
 	 * @param array       $params       Optionally extra parameters.
 	 *
-	 * @return Result|false
+	 * @return false
 	 * @since  2.0.0
 	 */
 	public function releaseVersion($version_id, $release_date = null, array $params = array())
@@ -858,7 +858,7 @@ class Api
 	{
 		return $this->api(self::REQUEST_GET, sprintf('/rest/api/2/project/%s/components', $project_key), array(), true);
 	}
-        
+
 	/**
 	 * Get all issue types with valid status values for a project.
 	 *
@@ -867,11 +867,11 @@ class Api
 	 * @return array
 	 * @since  2.0.0
 	 */
-	public function getProjectIssueTypes($project_key) 
+	public function getProjectIssueTypes($project_key)
 	{
 		return $this->api(self::REQUEST_GET, sprintf('/rest/api/2/project/%s/statuses', $project_key), array(), true);
 	}
-        
+
 	/**
 	 * Returns a list of all resolutions.
 	 *
@@ -880,7 +880,7 @@ class Api
 	 * @return array|false
 	 * @since  2.0.0
 	 */
-	public function getResolutions() 
+	public function getResolutions()
 	{
 		return $this->api(self::REQUEST_GET, '/rest/api/2/resolution', array(), true);
 	}
--- a/tests/Jira/ApiTest.php
+++ b/tests/Jira/ApiTest.php
@@ -4,7 +4,8 @@ namespace Tests\chobie\Jira;
 
 
 use chobie\Jira\Api;
-use chobie\Jira\Api\Authentication\Anonymous;
+use chobie\Jira\Api\Authentication\AuthenticationInterface;
+use Prophecy\Prophecy\ObjectProphecy;
 
 /**
  * Class ApiTest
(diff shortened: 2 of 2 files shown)


aik099:
@jpastoor , ready for review.

tests/Jira/ApiTest.php
```diff
@@ -14,116 +15,164 @@
...
+	 */
+	protected $api;
+
 	/**
-	 * Tests that any trailing slash in the endpoint url is removed before being stored in the object state
+	 * Credential.
+	 *
+	 * @var AuthenticationInterface
 	 */
-	public function testSetEndpointTrailingSlash()
```
jpastoor: You removed this test, please bring it back.

aik099: This test is replaced by `testSetendpoint` with data provider usage for cases, when trailing slash is present or not.

jpastoor: Yeah saw it, tried to remove the comment but you were too fast ;)

aik099: Then that's why I've got error from GitHub, when tried to reply to that
comment on website. :)

Decided to reply over email.

tests/Jira/ApiTest.php
```diff
@@ -14,116 +15,164 @@
...
-		$api = $this->getMockBuilder('\chobie\Jira\Api')->setMethods(array('api'))->disableOriginalConstructor()->getMock();
-		$api->expects($this->once())->method('api')->with(
-			$this->equalTo(Api::REQUEST_PUT),
-			$this->equalTo('/rest/api/2/version/111000'),
-			$this->equalTo($params)
+		$this->expectClientCall(
+			Api::REQUEST_PUT,
+			'/rest/api/2/version/111000',
+			$params,
+			''
```
jpastoor: Invalid return type, it would be either a Result or false.

tests/Jira/ApiTest.php
```diff
@@ -14,116 +15,164 @@
...
+			'nothing removed' => array('https://test.test', 'https://test.test'),
+		);
+	}
+
 	public function testUpdateVersion()
 	{
 		$params = array(
-			'released' => true,
-			'releaseDate' => '2010-07-06',
+			'overdue' => true,
```
jpastoor: Why change these inputs?

tests/Jira/ApiTest.php
```diff
@@ -14,116 +15,164 @@
...
+		parent::setUp();
 
-		// Make sure nothing is removed if there is no trailing slash
-		$url = 'https://urlwithouttrailing.slash';
-		$api->setEndPoint($url);
-		$this->assertEquals($url, $api->getEndpoint());
+		$this->credential = $this->prophesize('chobie\Jira\Api\Authentication\AuthenticationInterface')->reveal();
+		$this->client = $this->prophesize('chobie\Jira\Api\Client\ClientInterface');
+
+		$this->api = new Api(self::ENDPOINT, $this->credential, $this->client->reveal());
```
jpastoor: Indeed much nicer to mock the clientinterface instead of the Api::api method

src/Jira/Api.php
```diff
@@ -426,7 +426,7 @@ public function getVersions($project_key)
 	 * @param string $project_key Project Key.
 	 * @param string $name        The version name to match on.
 	 *
-	 * @return integer|null VersionId on match or null when there is no match.
+	 * @return array|null Version data on match or null when there is no match.
```
jpastoor: Nice catch

jpastoor:
Agree on the change to mock the ClientInterface instead of Api::api.

No fan of the change from phpunit to prophesy with the passive aggressive title suggesting that this is THE way to do stuff like this. With extracting the expectClientCall you add a layer of indirection, and the assertFalses on the outputs suggest that is a meaningful route to test but we don't do anything with that state.

jpastoor:
(OK for you to merge this one though since above is not important for me enough to be blocking or anything, just wanted to let you know the few things that crossed my mind :)

tests/Jira/ApiTest.php
aik099: Looking at previous test code it was obvious that sample data was copy-pasted from `releaseVersion` test, where exactly that data is submitted to JIRA.

In there we testing version update and that have nothing to do with indirect version release (it can be used to release if user wishes so, but that's not primary purpose of this method). To point that out even more I've used completely unrelated to releasing fields of the version.

tests/Jira/ApiTest.php
aik099: Is this comment addressed to the right line of code?

Ther `''` you've selected for comment is what `ClientInterface` implementing class returns. And it always returns strings, because that's what HTTP response is. It's `Api::api` that is json-decoding them and wrapping them into `Result` class.

jpastoor: Ah right

aik099:
> No fan of the change from phpunit to prophesy with the passive aggressive title suggesting that this is THE way to do stuff like this.

My apologies. I've renamed the PR. Actually `this is THE way` are thoughts of PHPUnit author (see https://thephp.cc/news/2015/02/phpunit-4-5-and-prophecy blogpost about why Prophecy is better mocking alternative as built-in PHPUnit-Mock library) .

> With extracting the expectClientCall you add a layer of indirection ...

This is because of default parameter values. When specifying expectation you don't have any auto-complete based on parameters of mocked method. Since mocked method have ~6 parameters, that must be specified in each expectation and in most cases we only provide 3 of them, then extracting in separate method reduces changes of getting things wrong, when writing unit tests. Also, at least for me, having nice `expect*` method is better than copy-pasting long expectation creating code all over the place.

>  and the assertFalses on the outputs suggest that is a meaningful route to test but we don't do anything with that state.

This tests, that return value of method stays as it is currently. Currently API calls, that are just successful, but doesn't return anything useful just return `false` to end user and that's what I test.

PR #131, opened 2016-08-18
Added client-side cache to Api::getResolutions 

### LINKED ISSUE #122: Improvements of "Api::getResolutions" method
The `Api::getResolutions` method (added in 2.0.0):

* have wrong DocBlock (the non-existing `$project_key` parameter)
* doesn't cache it's results in `Api` class as similar non-project/issue specific methods do (e.g. `getStatuses` and `getPriorities`)

jpastoor:
- Client side cache just like getStatuses, getPriorities etc
- Removed non-existing parameter in docblock
- Added unittest

Fixes #122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Minimal supported PHP version changed from 5.2 to 5.3 by [@chobie].
 - The `Api::getPriorties` renamed into `Api::getPriorities` (former method kept for BC reasons) by [@josevh].
 - Remove trailing slash from endpoint url by [@Procta].
-- Clearing local caches (statuses, priorities, fields and resolutions) on endpoint change [@jpastoor].
 - Added local cache to getResolutions [@jpastoor].
 
 ### Removed
@@ -36,6 +35,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed PHP deprecation notice, when creating issue attachments via `CurlClient` on PHP 5.5+ by [@DerMika].
 - The `Api::getRoles` call was always retuning an error by [@aik099].
 - Attempt to make a `DELETE` API call using `CurlClient` wasn't working by [@aik099].
+- Clearing local caches (statuses, priorities, fields and resolutions) on endpoint change [@jpastoor].
 
 ## [1.0.0] - 2014-07-27
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Minimal supported PHP version changed from 5.2 to 5.3 by [@chobie].
 - The `Api::getPriorties` renamed into `Api::getPriorities` (former method kept for BC reasons) by [@josevh].
 - Remove trailing slash from endpoint url by [@Procta].
+- Clearing local caches (statuses, priorities, fields and resolutions) on endpoint change [@jpastoor].
+- Added local cache to getResolutions [@jpastoor].
 
 ### Removed
 ...
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -68,30 +68,30 @@ class Api
 	protected $options = self::AUTOMAP_FIELDS;
 
 	/**
-	 * Client-side cache of Fields.
+	 * Client-side cache of Fields. List of fields when loaded, null when nothing is fetched yet.
 	 *
-	 * @var array|null List of fields when loaded, null when nothing is fetched yet
+	 * @var array|null
 	 */
 	protected $fields;
 
 	/**
-	 * Client-side cache of Priorities.
+	 * Client-side cache of Priorities. List of priorities when loaded, null when nothing is fetched yet.
 	 *
-	 * @var array|null List of priorities when loaded, null when nothing is fetched yet
+	 * @var array|null
 	 */
 	protected $priorities;
 
 	/**
-	 * Client-side cache of Statuses.
+	 * Client-side cache of Statuses. List of statuses when loaded, null when nothing is fetched yet.
 	 *
-	 * @var array|null List of statuses when loaded, null when nothing is fetched yet
+	 * @var array|null
 	 */
 	protected $statuses;
 
 	/**
-	 * Client-side cache of Resolutions.
+	 * Client-side cache of Resolutions. List of resolutions when loaded, null when nothing is fetched yet.
 	 *
-	 * @var array|null List of resolutions when loaded, null when nothing is fetched yet
+	 * @var array|null
 	 */
 	protected $resolutions;
 
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -70,28 +70,28 @@ class Api
 	/**
 	 * Client-side cache of Fields.
 	 *
-	 * @var array
+	 * @var array|null List of fields when loaded, null when nothing is fetched yet
 	 */
 	protected $fields;
 
 	/**
 	 * Client-side cache of Priorities.
 	 *
-	 * @var array
+	 * @var array|null List of priorities when loaded, null when nothing is fetched yet
 	 */
 	protected $priorities;
 
 	/**
 	 * Client-side cache of Statuses.
 	 *
-	 * @var array
+	 * @var array|null List of statuses when loaded, null when nothing is fetched yet
 	 */
 	protected $statuses;
 
 	/**
 	 * Client-side cache of Resolutions.
 	 *
-	 * @var array
+	 * @var array|null List of resolutions when loaded, null when nothing is fetched yet
 	 */
 	protected $resolutions;
 
@@ -148,12 +148,26 @@ class Api
 	 */
 	public function setEndpoint($url)
 	{
-		$this->fields = array();
-
 		// Remove trailing slash in the url.
 		$url = rtrim($url, '/');
 
-		$this->endpoint = $url;
+		if ( $url != $this->endpoint ) {
+			$this->endpoint = $url;
+			$this->clearLocalCaches();
+		}
+	}
+
+	/**
+	 * Helper method to clear the local caches. Is called when switching endpoints
+	 *
+	 * @return void
+	 */
+	protected function clearLocalCaches()
+	{
+		$this->fields = null;
+		$this->priorities = null;
+		$this->statuses = null;
+		$this->resolutions = null;
 	}
 
 	/**
@@ -163,13 +177,14 @@ class Api
 	 */
 	public function getFields()
 	{
-		if ( !count($this->fields) ) {
+		// Fetch fields when the method is called for the first time.
+		if ( $this->fields === null ) {
 			$fields = array();
-			$_fields = $this->api(self::REQUEST_GET, '/rest/api/2/field', array());
+			$result = $this->api(self::REQUEST_GET, '/rest/api/2/field', array(), true);
 
 			/* set hash key as custom field id */
-			foreach ( $_fields->getResult() as $k => $v ) {
-				$fields[$v['id']] = $v;
+			foreach ( $result as $field ) {
+				$fields[$field['id']] = $field;
 			}
 
 			$this->fields = $fields;
@@ -463,13 +478,14 @@ class Api
 	 */
 	public function getPriorities()
 	{
-		if ( !count($this->priorities) ) {
+		// Fetch priorities when the method is called for the first time.
+		if ( $this->priorities === null ) {
 			$priorities = array();
-			$result = $this->api(self::REQUEST_GET, '/rest/api/2/priority', array());
+			$result = $this->api(self::REQUEST_GET, '/rest/api/2/priority', array(), true);
 
 			/* set hash key as custom field id */
-			foreach ( $result->getResult() as $k => $v ) {
-				$priorities[$v['id']] = $v;
+			foreach ( $result as $priority ) {
+				$priorities[$priority['id']] = $priority;
 			}
 
 			$this->priorities = $priorities;
@@ -498,13 +514,14 @@ class Api
 	 */
 	public function getStatuses()
 	{
-		if ( !count($this->statuses) ) {
+		// Fetch statuses when the method is called for the first time.
+		if ( $this->statuses === null ) {
 			$statuses = array();
-			$result = $this->api(self::REQUEST_GET, '/rest/api/2/status', array());
+			$result = $this->api(self::REQUEST_GET, '/rest/api/2/status', array(), true);
 
 			/* set hash key as custom field id */
-			foreach ( $result->getResult() as $k => $v ) {
-				$statuses[$v['id']] = $v;
+			foreach ( $result as $status ) {
+				$statuses[$status['id']] = $status;
 			}
 
 			$this->statuses = $statuses;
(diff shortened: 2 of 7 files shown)


src/Jira/Api.php
```diff
@@ -875,14 +882,23 @@ public function getProjectIssueTypes($project_key)
...
 	 * @since  2.0.0
 	 */
 	public function getResolutions()
 	{
-		return $this->api(self::REQUEST_GET, '/rest/api/2/resolution', array(), true);
+		if ( !count($this->resolutions) ) {
+			$resolutions = array();
+			$result = $this->api(self::REQUEST_GET, '/rest/api/2/resolution', array());
+
+			foreach ( $result->getResult() as $k => $v ) {
```
aik099: Please do this:

1. remove `$k` variable, because array key isn't being used
2. rename `$v` variable into `$resolution_data`

src/Jira/Api.php
```diff
@@ -875,14 +882,23 @@ public function getProjectIssueTypes($project_key)
...
-	 * @param string $project_key Project key.
-	 *
-	 * @return array|false
+	 * @return array
 	 * @since  2.0.0
 	 */
 	public function getResolutions()
 	{
-		return $this->api(self::REQUEST_GET, '/rest/api/2/resolution', array(), true);
+		if ( !count($this->resolutions) ) {
```
aik099: Considering that default value of `resolutions` property is `null` instead of empty array it's not good to treat it as array and try to count values in it. Maybe it even results in a notice.

I think, that assigning an empty array to `resolutions` property would be correct solution here.

P.S.

If you wish you can do the same with `fields`, `priorities` and `statuses` properties, that are used the same way.

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,40 @@ public function testFindVersionByName()
 		);
 	}
 
+	public function testGetResolutions()
+	{
+		$response = '[{"self":"https://test.atlassian.net/rest/api/2/resolution/1","id":"1","description":"A fix for this issue is checked into the tree and tested.","name":"Fixed"},{"self":"https://test.atlassian.net/rest/api/2/resolution/10000","id":"10000","description":"This issue won\'t be actioned.","name":"Won\'t Do"}]';
```
aik099: Please define it as PHP array here.

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,40 @@ public function testFindVersionByName()
...
+
+		$this->expectClientCall(
+			Api::REQUEST_GET,
+			'/rest/api/2/resolution',
+			array(),
+			$response
```
aik099: Use `json_encode($response)` here.

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,40 @@ public function testFindVersionByName()
...
+		);
+
+		$actual = $this->api->getResolutions();
+
+		$this->assertEquals(array(
+			'1' => array(
```
aik099: Actual array content can be replaced with `$response[0]` to more explicitly state, that untouched array from API response is returned.

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,40 @@ public function testFindVersionByName()
...
+				'self' => 'https://test.atlassian.net/rest/api/2/resolution/1',
+				'id' => '1',
+				'description' => 'A fix for this issue is checked into the tree and tested.',
+				'name' => 'Fixed',
+			),
+			'10000' => array(
```
aik099: Actual array content can be replaced with `$response[1]` to more explicitly state, that untouched array from API response is returned.

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,40 @@ public function testFindVersionByName()
...
+				'id' => '10000',
+				'description' => 'This issue won\'t be actioned.',
+				'name' => 'Won\'t Do',
+			),
+		), $actual);
+
+		// Second time we call the method the results should be cached and not trigger an API Request.
+		$this->client->sendRequest(Api::REQUEST_GET, '/rest/api/2/resolution', array(), self::ENDPOINT, $this->credential)
+			->shouldNotBeCalled();
+		$this->api->getResolutions();
```
aik099: You also need to assert here, that return value is correct.

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,40 @@ public function testFindVersionByName()
...
+			'10000' => array(
+				'self' => 'https://test.atlassian.net/rest/api/2/resolution/10000',
+				'id' => '10000',
+				'description' => 'This issue won\'t be actioned.',
+				'name' => 'Won\'t Do',
+			),
+		), $actual);
+
+		// Second time we call the method the results should be cached and not trigger an API Request.
+		$this->client->sendRequest(Api::REQUEST_GET, '/rest/api/2/resolution', array(), self::ENDPOINT, $this->credential)
```
aik099: Have you tried to remove caching layer in `getResolutions` method and checked that test will failed because of this assertion?

src/Jira/Api.php
jpastoor: Agreed that it is a bit wonky now. Refactored it to use null when the list is not fetched, and doa  type safe check on that. 

This way we can make a distinction between non-fetched and empty list. (Internally only, consumers always get an array returned when using the method)

tests/Jira/ApiTest.php
jpastoor: Yep that works

tests/Jira/ApiTest.php
jpastoor: Guess thats a preference of style. I'd done it like this to more explicitly model what goes in and out.

jpastoor: Think I found a nice compromis between our solutions. Keeps the tests nice and short, no duplication but also the raw JSON to your disposal if needed:

		$response = file_get_contents(__DIR__.'/resources/api_status.json');

		$this->expectClientCall(
			Api::REQUEST_GET,
			'/rest/api/2/resolution',
			array(),
			$response
		);

		$actual = $this->api->getResolutions();

		$response_decoded = json_decode($response, true);

		$expected = [
			'1' => $response_decoded[0],
			'3' => $response_decoded[1],
		];
		$this->assertEquals($expected, $actual);

Also writing the same tests for getFields, getStatuses, getPriorities

src/Jira/Api.php
aik099: I guess you haven't pushed that change.

tests/Jira/ApiTest.php
aik099: Looks almost good:

1.  why you've called file `api_status.json` while it is returning resolutions, not statuses?
2. maybe we should place `resources` folder right in `/tests/` and not based on test class location?

src/Jira/Api.php
aik099: I usually check such things as `!isset($this->resolutions)`. This way we indeed can handle the case, when `JIRA instance have 0 resolutions defined` VS case, when `we haven't queried resolutions yet`.

src/Jira/Api.php
```diff
@@ -68,27 +68,34 @@ class Api
 	protected $options = self::AUTOMAP_FIELDS;
 
 	/**
-	 * Fields.
+	 * Client-side cache of Fields.
 	 *
-	 * @var array
+	 * @var array|null List of fields when loaded, null when nothing is fetched yet
```
aik099: In multi-line notation of `@var` DocBlock the description should go in dedicated place (up). In single line `@var` DocBlock notation indeed the description would be written right after type name.

src/Jira/Api.php
```diff
@@ -68,27 +68,34 @@ class Api
...
 	 */
 	protected $fields;
 
 	/**
-	 * Priorities.
+	 * Client-side cache of Priorities.
 	 *
-	 * @var array
+	 * @var array|null List of priorities when loaded, null when nothing is fetched yet
```
aik099: In multi-line notation of `@var` DocBlock the description should go in dedicated place (up). In single line `@var` DocBlock notation indeed the description would be written right after type name.

src/Jira/Api.php
```diff
@@ -68,27 +68,34 @@ class Api
...
 	 */
 	protected $priorities;
 
 	/**
-	 * Statuses.
+	 * Client-side cache of Statuses.
 	 *
-	 * @var array
+	 * @var array|null List of statuses when loaded, null when nothing is fetched yet
```
aik099: In multi-line notation of `@var` DocBlock the description should go in dedicated place (up). In single line `@var` DocBlock notation indeed the description would be written right after type name.

src/Jira/Api.php
```diff
@@ -68,27 +68,34 @@ class Api
...
 	 */
 	protected $statuses;
 
 	/**
+	 * Client-side cache of Resolutions.
+	 *
+	 * @var array|null List of resolutions when loaded, null when nothing is fetched yet
```
aik099: In multi-line notation of `@var` DocBlock the description should go in dedicated place (up). In single line `@var` DocBlock notation indeed the description would be written right after type name.

src/Jira/Api.php
```diff
@@ -141,12 +148,26 @@ public function getEndpoint()
 	 */
 	public function setEndpoint($url)
 	{
-		$this->fields = array();
-
 		// Remove trailing slash in the url.
 		$url = rtrim($url, '/');
 
-		$this->endpoint = $url;
+		if ( $url != $this->endpoint ) {
```
aik099: Use `!==` here.

src/Jira/Api.php
```diff
@@ -156,13 +177,14 @@ public function setEndpoint($url)
 	 */
 	public function getFields()
 	{
-		if ( !count($this->fields) ) {
+		// Fetch fields when the method is called for the first time.
+		if ( $this->fields === null ) {
```
aik099: I saw this approach in Symfony, but wouldn't `isset($this->fields)` be more explicit (I've wrote about this in another comment)?

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,114 @@ public function testFindVersionByName()
...
+		$this->assertEquals($expected, $this->api->getFields(), 'Calling twice did not yield the same results');
+	}
+
+	public function testGetStatuses()
+	{
+		$response = file_get_contents(__DIR__ . '/resources/api_status.json');
+
+		$this->expectClientCall(
+			Api::REQUEST_GET,
+			'/rest/api/2/resolution',
```
aik099: I wonder how this test pass, when you make `resolution` call to get status?

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,114 @@ public function testFindVersionByName()
...
+			array(),
+			$response
+		);
+
+		$actual = $this->api->getResolutions();
```
aik099: Ha, probably `->getStatuses()` should be in here. Better to check that test coverage is correct in your IDE.

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,114 @@ public function testFindVersionByName()
...
+		$response_decoded = json_decode($response, true);
+
+		$expected = array(
+			'1' => $response_decoded[0],
+			'3' => $response_decoded[1],
+		);
+		$this->assertEquals($expected, $actual);
+
+		// Second time we call the method the results should be cached and not trigger an API Request.
+		$this->client->sendRequest(Api::REQUEST_GET, '/rest/api/2/resolution', array(), self::ENDPOINT, $this->credential)
```
aik099: I wonder how this test pass, when you make `resolution` call to get status?

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,114 @@ public function testFindVersionByName()
...
+			->shouldNotBeCalled();
+		$this->assertEquals($expected, $this->api->getResolutions(), 'Calling twice did not yield the same results');
```
aik099: Ha, probably `->getStatuses()` should be in here. Better to check that test coverage is correct in your IDE.

tests/Jira/ApiTest.php
```diff
@@ -149,6 +149,114 @@ public function testFindVersionByName()
 		);
 	}
 
+	public function testGetResolutions()
```
aik099: All these tests look almost the same. Proposing to create one universal test with a data provider, that would accept these parameters:

* json filename (e.g. `api_resolution.json`)
* API url (e.g. `/rest/api/2/resolution`)
* expected result (maybe we can make resolutions/fields/statuses/etc. have same IDs and then we don't need to change `$expected` variable at all

aik099:
PR branch was deleted. Are you force pushing maybe?

src/Jira/Api.php
jpastoor: I tend to use strict type comparison instead of the more magicky methods like empty() and isset(). Matter of preference I guess. (empty("0") == true or false? :D)

tests/Jira/ApiTest.php
jpastoor: Nice catch, will fix

src/Jira/Api.php
aik099: No worries. Just use it consistently so we don't me different null checking styles in different places.

Yeah, the `empty` function is hell. I never use it.

src/Jira/Api.php
```diff
@@ -68,27 +68,34 @@ class Api
 	protected $options = self::AUTOMAP_FIELDS;
 
 	/**
-	 * Fields.
+	 * Client-side cache of Fields. List of fields when loaded, null when nothing is fetched yet.
```
aik099: The `Fields` should be in lowercase. It was capitalized because it was 1st word in sentence.

src/Jira/Api.php
```diff
@@ -68,27 +68,34 @@ class Api
...
 	 *
-	 * @var array
+	 * @var array|null
 	 */
 	protected $fields;
 
 	/**
-	 * Priorities.
+	 * Client-side cache of Priorities. List of priorities when loaded, null when nothing is fetched yet.
```
aik099: The `Priorities` should be in lowercase. It was capitalized because it was 1st word in sentence.

src/Jira/Api.php
```diff
@@ -68,27 +68,34 @@ class Api
...
 	 *
-	 * @var array
+	 * @var array|null
 	 */
 	protected $priorities;
 
 	/**
-	 * Statuses.
+	 * Client-side cache of Statuses. List of statuses when loaded, null when nothing is fetched yet.
```
aik099: The `Statuses` should be in lowercase. It was capitalized because it was 1st word in sentence.

src/Jira/Api.php
```diff
@@ -68,27 +68,34 @@ class Api
...
 	 *
-	 * @var array
+	 * @var array|null
 	 */
 	protected $statuses;
 
 	/**
+	 * Client-side cache of Resolutions. List of resolutions when loaded, null when nothing is fetched yet.
```
aik099: The `Resolutions` should be in lowercase. It was capitalized because it was 1st word in sentence.

aik099:
Code looks good. Just noticed, that `CHANGELOG.md` wasn't updated to mention:

1. change to `Api::getResolutions` method to support caching
2. reset of local caches on endpoint url change for existing `Api` class instance

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Minimal supported PHP version changed from 5.2 to 5.3 by [@chobie].
 - The `Api::getPriorties` renamed into `Api::getPriorities` (former method kept for BC reasons) by [@josevh].
 - Remove trailing slash from endpoint url by [@Procta].
+- Clearing local caches (statuses, priorities, fields and resolutions) on endpoint change [@jpastoor].
```
aik099: This is a bugfix and needs to be moved to `Fixes` section.

aik099:
Merging. Thanks, @jpastoor .

Issue #69, opened 2016-02-28
Suggestion to remove support for PHP 5.3 for 2.0 release


jpastoor:
I think the 2.0 release would also be a nice moment to stop support for PHP 5.3. It's 6.5 years old and it limits our use of short notations, traits and other libs (like newer version of PHPUnit) that have dropped support for 5.3 as well.

Are there people who would have problems with that?

aik099:
Not sure that mentioned PHP features added in PHP versions after 5.3 would make any difference in the library currently. According to https://seld.be/notes/php-versions-stats-2015-edition the 10% of Web Servers are still using PHP 5.3 and maybe some of them are existing/potential library users.

What we can do instead is incrementing required PHP version to 5.3.7 where bcrypt fix was made.

DerMika:
I have an application using php5.4 with this library, so as long as you keep supporting 5.4 forba while, I'm good

addersuk:
With 5.4 already at EOL and 5.5 in security support only, i would definately recommend dropping 5.3 and even 5.4. 
If 2.0 is a new release you should set suitable requirements as all recently released frameworks require 5.5 or later, therefore i can see no reason why this project shouldn't follow these support levels.  Support of older versions requires additional burden, and as the existing library works, users of older versions can continue to use v1.0.

DerMika:
OK, but will 1.0 continue to be supported?

aik099:
@DerMika , we can't increase PHP version requirement for 1.0 release, so yes it will support PHP 5.3, but won't receive new features as 2.0.0 will do.

aik099:
Supporting PHP 5.3 doesn't require extra resources and that's why it doesn't make sense not to support it.

jpastoor:
Still think its a good idea. The people running old and EOL PHP versions are most likely not our customer base doing fancy stuff with composer/github.

aik099:
We're not the ones to make decisions about that. As I said before: if we're not suffering from supporting EOL PHP versions, then no logical reason to stop supporting them just because they are EOL.

PHP doesn't force developers to install newer PHP versions and for that reason I think it's best to leave decision in developers hands about what PHP version should be used.

aik099:
The minimal supported PHP version for v2.0.0 release is PHP 5.6.

PR #133, opened 2016-08-18
Added phpcs to composer and prefer stable releases of dependencies

jpastoor:
**PHPCS**
Since we require devs to run the coding standard we might as well make it as easy as possible by adding it to the require-dev dependencies, just like we do with phpunit. Updated the CONTRIBUTING.md to reflect this.

**minimum-stability**
Also I think it's bad practise to se the minimum-stability setting of all your dependencies to dev, since then we also get the dev builds of dependencies that have tested, stable releases. This change ensures the latest master of the CodingStyle, but the latest stable release of other libs.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^4.1|^5.0",
-        "aik099/coding-standard": "dev-master@dev",
+        "aik099/coding-standard": "dev-master",
         "squizlabs/php_codesniffer": "^2.6"
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -11,11 +11,6 @@
             "email": "chobieeee@php.net"
         }
     ],
-    "config": {
-        "platform": {
-            "php": "5.3.29"
-        }
-    },
     "require": {
         "php": ">=5.3.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -11,13 +11,13 @@
             "email": "chobieeee@php.net"
         }
     ],
-    "minimum-stability": "dev",
     "require": {
         "php": ">=5.3.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.1|^5.0",
-        "aik099/coding-standard": "dev-master"
+        "aik099/coding-standard": "dev-master@dev",
+        "squizlabs/php_codesniffer": "^2.6"
     },
     "autoload": {
         "psr-4" : {
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,11 @@
             "email": "chobieeee@php.net"
         }
     ],
+    "config": {
+        "platform": {
+            "php": "5.3.29"
+        }
+    },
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -11,13 +11,13 @@
...
         }
     ],
-    "minimum-stability": "dev",
     "require": {
         "php": ">=5.3.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.1|^5.0",
-        "aik099/coding-standard": "dev-master"
+        "aik099/coding-standard": "dev-master@dev",
```
aik099: Is `@dev` needed for stuff to work?

jpastoor:
I think it is needed when using branches in the "require" section if your parent project has strict stability settings. Does not seem to be needed for require-dev though.

aik099:
The `require-dev` only work in topmost `composer.json` (see https://getcomposer.org/doc/04-schema.md#require-dev). This way we can do anything in here and it won't affect project that uses this library as a dependency.

aik099:
The build is failing. I think you've specified PHP_CodeSniffer, that can't work on PHP 5.3.

jpastoor:
No  the problem was my composer update on a PHP 5.6 distro, causing it to fetch phpunit of the 5.x branch. Fixed it with a directive that tells composer that the target system is a 5.3. Since dependencies are dropping 5.3 support as well I think #69 is still very relevant.

Anyways this is fixed now and can be merged I think.

Another solution might be to update the travis directives to use "composer update" instead of a "composer install". Composer can then pick the dependencies for its own PHP version. Maybe checking in the composer.lock is not wishful at all when supporting so many PHP versions.

aik099:
> Another solution might be to update the travis directives to use "composer update" instead of a "composer install". Composer can then pick the dependencies for its own PHP version. Maybe checking in the composer.lock is not wishful at all when supporting so many PHP versions.

This would end up in fancy Travis CI composer cache not being used and 3x slower build. That is from what we started and later I've switched to dependencies from lock file (-1 remote query from Composer side) and cache all that composer was downloaded.

composer.json
```diff
@@ -11,13 +11,18 @@
             "email": "chobieeee@php.net"
         }
     ],
-    "minimum-stability": "dev",
+    "config": {
+        "platform": {
+            "php": "5.3.29"
```
aik099: Won't this affect library users having newer PHP versions, e.g. by always downloading PHP 5.3 compatible dev dependencies?

aik099: I have all PHP versions installed locally. If you'd like I can send diff to you with correct `composer.lock` file for your `composer.json` (without that `config` setting)?

aik099:
Here are `composer.json` and `composer.lock` files, that should work on PHP 5.3 without extra changes to `composer.json`:

[composer_files.zip](https://github.com/chobie/jira-api-restclient/files/425042/composer_files.zip)

aik099:
Merging, thanks @jpastoor .

Issue #73, opened 2016-02-28
Inconsistency in return type of API calls 


jpastoor:
The API::api() method has a parameter called $return_as_json.

Currently I see two problems
1. When set to true, data is not returned as JSON but as associative array instead.
2. The API class has many helper methods that wrap around the api() method. About half of them return as associative array, the others as Result. 

I haven't looked into this too deeply yet but it might be nice to 
- standardize the wrapper methods to one return type
- and/or propogate the return_as_json by adding that option in the wrapper method parameter list as well
- maybe rename return_as_json to return_as_array

aik099:
Having `api` method on `Api` class is confusing enough as it is :smile: 

Looking at `Api::api` method the `$return_as_json` in fact means:

1. get data
2. parse it into array using `json_decode`
3. return it

Without this `$return_as_json` parameter the same array as above is wrapped with `Result` class, which is some kind of iterator creation attempt, that walks over queried issues.

Maybe doing a rename on that parameter will solve that problem. We can't really introduce `apiJSON` method to solve it.

aik099:
After pondering about this for a while I have some usable responses now:

> When set to true, data is not returned as JSON but as associative array instead.
> The API class has many helper methods that wrap around the api() method. About half of them return as associative array, the others as Result.

> standardize the wrapper methods to one return type

👍 

The `Result` class obviously is tailored for representing issue list that `Walker` can iterate upon. For all other response types (e.g. project versions, statuses, etc.) the only `Result::getResult` method would return underlying array as-as.

Since it's 2.0.0 we can:

* change default value for `$return_as_json` to `true` (from `false`)
* in place (1 or 2 places most likely), where we're getting issue list back specify `$return_as_json` as `false`

> and/or propogate the return_as_json by adding that option in the wrapper method parameter list as well

@jpastoor , what do you mean?

> maybe rename return_as_json to return_as_array

I've created #119 with exactly that proposition.

jpastoor:
I'll write a PR for this tomorrow

PR #134, opened 2016-08-18
Renamed Api::api() parameter $return_as_json to $return_as_array

### LINKED ISSUE #119: Rename "$return_as_json" into "$return_as_array"
Currently `Api::api` method has `$return_as_json` parameter which:

* when set to `true` would return **an array**
* when set to `false` would return **the `Result` class object**

No JSON involved in either of returned values.

Proposing to rename parameter to `$return_as_array` because it at least will be true and update parameter DocBlock comment.

jpastoor:
Fixes #119  
Part of #73

Separating the easy from the possible controversial part.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - The `Api::getPriorties` renamed into `Api::getPriorities` by [@josevh].
 - Remove trailing slash from endpoint url by [@Procta].
 - Added local cache to getResolutions [@jpastoor].
+- Renamed Api::api() parameter $return_as_json to $return_as_array [@jpastoor].
 
 ### Removed
 ...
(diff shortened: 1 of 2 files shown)


src/Jira/Api.php
```diff
@@ -704,7 +704,7 @@ public function createRemotelink(
 	 * @param string       $method         Request method.
 	 * @param string       $url            URL.
 	 * @param array|string $data           Data.
-	 * @param boolean      $return_as_json Return results as JSON.
+	 * @param boolean      $return_as_array Return results as JSON.
```
aik099: From GitHub UI looks like whole DocBlock needs to be reformatted (Ctrl+Alt+L in PhpStorm, when DocBlock is selected) to reflect new length of parameter name.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - The `Api::getPriorties` renamed into `Api::getPriorities` by [@josevh].
 - Remove trailing slash from endpoint url by [@Procta].
 - Added local cache to getResolutions [@jpastoor].
+- Renamed Api::api() parameter $return_as_json to $return_as_array [@jpastoor].
```
aik099: The `by` is missing before username (in other places `some text by [@username]`). I've noticed it now, but PRs merged today also have that issue.

aik099:
Merging, thanks @jpastoor .

PR #145, opened 2016-09-13
Added option to override the filename when creating attachments

### LINKED ISSUE #141: Add option to override the name of an attached file
It would be useful to be able to set the name of a file that is uploaded manually instead of relying on the path.

We have a form that people use to create issues in JIRA and to implement the file upload we have to move the file to a temp location with it's real name before submitting to JIRA. I'd like to be able to do this without having to move the file.

I can do a PR for this but wanted to get an opinion first.

betterphp:
Closes #141

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Api::getProjectIssueTypes` call for getting project issue types and issue statuses connected to them by [@N-M].
 - Added `Api::getResolutions` call for getting available issue resolutions by [@N-M].
 - Allow configuring issues queried per page in `Walker` class by [@aik099].
-- Added optional override for the filename in Api::createAttachment by [@betterphp]
+- Added optional override for the filename in `Api::createAttachment` by [@betterphp]
 
 ### Changed
 - Classes/interfaces were renamed to use namespaces by [@chobie].
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Api::getProjectIssueTypes` call for getting project issue types and issue statuses connected to them by [@N-M].
 - Added `Api::getResolutions` call for getting available issue resolutions by [@N-M].
 - Allow configuring issues queried per page in `Walker` class by [@aik099].
+- Added optional override for the filename in Api::createAttachment by [@betterphp]
 
 ### Changed
 - Classes/interfaces were renamed to use namespaces by [@chobie].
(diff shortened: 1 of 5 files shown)


src/Jira/Api/Client/CurlClient.php
```diff
@@ -159,16 +159,21 @@ public function sendRequest(
 	 * If necessary, replace curl file @ string with a CURLFile object (for PHP 5.5 and up)
 	 *
 	 * @param string $file_string The string in @-format as it is used on PHP 5.4 and older.
+	 * @param string $name Optional override for the filename
 	 *
 	 * @return \CURLFile|string
 	 */
-	protected function getCurlValue($file_string)
+	protected function getCurlValue($file_string, $name)
```
aik099: Please change `$name` into `$name = null`.

src/Jira/Api/Client/CurlClient.php
```diff
@@ -159,16 +159,21 @@ public function sendRequest(
...
 	{
+		if ($name === null){
```
aik099: CS: incorrect spacing.

src/Jira/Api.php
```diff
@@ -644,17 +644,15 @@ public function releaseVersion($version_id, $release_date = null, array $params
 	 *
 	 * @param string $issue_key Issue key.
 	 * @param string $filename  Filename.
-	 * @param array  $options   Options.
+	 * @param string $name      Name.
```
aik099: CS: Probably incorrectly aligned DocBlock.

betterphp: I can't really see what's wrong here :/

betterphp:
When I run phpcs it reports about 40 errors so I assumed it was not really used. Could that be because of my local PHP version?

betterphp:
Also sorry for the delay, I'm doing this for work and have been moved on to another project for a while

src/Jira/Api.php
aik099: Maybe it's just GitHub replaced all TABs with SPACEs and it look wrong 😉

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `Api::getProjectIssueTypes` call for getting project issue types and issue statuses connected to them by [@N-M].
 - Added `Api::getResolutions` call for getting available issue resolutions by [@N-M].
 - Allow configuring issues queried per page in `Walker` class by [@aik099].
+- Added optional override for the filename in Api::createAttachment by [@betterphp]
```
aik099: Not done.

aik099:
I've run PHP_CodeSniffer only inspection on files changed in PR and it reported 7 issues and none were in the changed lines. It's also important to use correding PHPCS standard (see contribution instructions) and not default PEAR one.

aik099:
Merging, thanks @betterphp .

PR #147, opened 2016-10-08
Disable CURLOPT_HTTP_VERSION override

benPesso:
Forcing CURLOPT_HTTP_VERSION to CURL_HTTP_VERSION_1_0 results in "SSLRead() return error -9806" error when connecting making a request to JIRA.

### DIFF
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -93,7 +93,6 @@ class CurlClient implements ClientInterface
 			curl_setopt($curl, CURLOPT_USERPWD, sprintf('%s:%s', $credential->getId(), $credential->getPassword()));
 		}
 
-// 		curl_setopt($curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_0);
 		curl_setopt($curl, CURLOPT_SSL_VERIFYPEER, false);
 		curl_setopt($curl, CURLOPT_SSL_VERIFYHOST, 0);
 		curl_setopt($curl, CURLOPT_VERBOSE, $debug);
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -93,7 +93,7 @@ class CurlClient implements ClientInterface
 			curl_setopt($curl, CURLOPT_USERPWD, sprintf('%s:%s', $credential->getId(), $credential->getPassword()));
 		}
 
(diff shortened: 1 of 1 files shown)


aik099:
It does not. Tests doesn't fail as well.

Maybe this happens on specific JIRA + PHP version only?

benPesso:
Could be. Running with built-in PHP on MacOS Sierra (PHP 5.5). JIRA is v7.0.9.

aik099:
1. what does `php -r "print_r(openssl_get_cert_locations());"` when you run it in Terminal?
2. do you have `/usr/local/libressl/etc` folder and what's in it.

When I first upgraded to macOS Sierra (from Yosemite) I've got that folder, but it was empty. However executing 1st command showed that OpenSSL version compiled on macOS was using that folder. That resulted in all SSL requests made from PHP failing due missing PEM file to verify validity of issues of SSL certificates used by websites you're connecting to from PHP.

benPesso:
That method is only available from 5.6, but either way, I don't have the `libressl` folder, so it's likely I'm missing the certificates.

src/Jira/Api/Client/CurlClient.php
```diff
@@ -93,7 +93,7 @@ public function sendRequest(
 			curl_setopt($curl, CURLOPT_USERPWD, sprintf('%s:%s', $credential->getId(), $credential->getPassword()));
 		}
 
-		curl_setopt($curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_0);
+// 		curl_setopt($curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_0);
```
aik099: Just remove this line.

aik099:
The curl documentation page recommends not to use this option at all and I haven't found any reason to use it to work with JIRA. That's I'm OK with it's removal.

Just please remove it completely and not comment-out. All tests pass.

aik099:
Actually you can also drop a line in the `CHANGELOG.md` file (the `Fixed` section for `Unreleased` version) about this change if you think it's worthy. Note, that you then also need to add mapping between your GitHub username and your GitHub profile link at the bottom of the file as well.

benPesso:
Thanks, but it's a really minor change. I'm ok with not having the attribution. :)

aik099:
Merged, thanks @benPesso .

benPesso:
Pleasure, @aik099.

PR #149, opened 2016-11-13
Allow counting queried issues of "Walker" class

### LINKED ISSUE #117: Allow issue walker to be countable
So far in all my projects I've used this package via Walker:

``` php
$jira = jiraApiClient();
$jiraIssues = new Walker($jira);

$jiraIssues->push('project = EXAMPLE');

foreach($jiraIssues as $issue) {
          $IssueSummary = $issue->getFields()['Summary'];
}
```

However using it this way I can't use any of the methods I see in the Result.php, like `getTotal()` etc. Ideally I'd like to be able to do something like `$jiraIssues->getTotal()` before processing the results in the foreach.

How can I use these methods? Is there some best practice to do what I'm trying to?

aik099:
Closes #117

### DIFF
--- a/tests/Jira/Issues/WalkerTest.php
+++ b/tests/Jira/Issues/WalkerTest.php
@@ -238,6 +238,33 @@ class WalkerTest extends \PHPUnit_Framework_TestCase
 		);
 	}
 
+	public function testCounting()
+	{
+		// Full 1st page.
+		$search_response1 = $this->generateSearchResponse('PRJ1', 5, 7);
+		$this->api->search('test jql', 0, 5, 'description')->willReturn($search_response1);
+
+		// Incomplete 2nd page.
+		$search_response2 = $this->generateSearchResponse('PRJ2', 2, 7);
+		$this->api->search('test jql', 5, 5, 'description')->willReturn($search_response2);
+
+		$walker = $this->createWalker(5);
+		$walker->push('test jql', 'description');
+
+		$this->assertEquals(7, count($walker));
+
+		$found_issues = array();
+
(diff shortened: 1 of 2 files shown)


tests/Jira/Issues/WalkerTest.php
```diff
@@ -238,6 +238,33 @@ public function testIssuesPassedThroughDelegate()
...
+		$this->api->search('test jql', 0, 5, 'description')->willReturn($search_response1);
+
+		// Incomplete 2nd page.
+		$search_response2 = $this->generateSearchResponse('PRJ2', 2, 7);
+		$this->api->search('test jql', 5, 5, 'description')->willReturn($search_response2);
+
+		$walker = $this->createWalker(5);
+		$walker->push('test jql', 'description');
+
+		$this->assertEquals(7, count($walker));
```
aik099: Counting would involve querying 1st page of issues and the `foreach` after it ensures, that this query won't loose all issues on 1st page as a result.

aik099:
@jpastoor , looks good?

jpastoor:
Looks good to me.
Personally I'm more of a fan of using a NULL value instead of -1 when denoting a not determined number. (more clearer and works also when negative numbers are valid entries as well). But feel free to merge it either way, it's a class internal anyways.

aik099:
> Personally I'm more of a fan of using a NULL value instead of -1 when denoting a not determined number.

Will fix that. I wasn't sure about `-1` either, but since property is of integer type I've used `-1`.

aik099:
Replaced `-1` with `null` as requested.

PR #158, opened 2017-07-26
Update README with generic API call usage

jdecool:


README.md
```diff
@@ -30,6 +31,12 @@ foreach ( $walker as $issue ) {
     var_dump($issue);
     // Send custom notification here.
 }
+
+// generic API call
```
aik099: Please format example so that it will fit on single screen without wrapping. The multi-line function call will do just fine.

jdecool:
I close this PR because I've deleted the initial fork and I'm not able to update this PR.

I will send an another PR in few days.

PR #160, opened 2017-09-12
Set timeout for fail gracefully

ish1301:
Sometime JIRA server is down/moved and API takes forever to respond. So consistent timeout address this issue and fail with message that "JIRA server have timed out after 30000 mili seconds"

### DIFF
--- a/src/Jira/Api/Client/CurlClient.php
+++ b/src/Jira/Api/Client/CurlClient.php
@@ -93,6 +93,7 @@ class CurlClient implements ClientInterface
 			curl_setopt($curl, CURLOPT_USERPWD, sprintf('%s:%s', $credential->getId(), $credential->getPassword()));
 		}
 
+		curl_setopt($curl, CURLOPT_TIMEOUT, 30);
 		curl_setopt($curl, CURLOPT_SSL_VERIFYPEER, false);
 		curl_setopt($curl, CURLOPT_SSL_VERIFYHOST, 0);
 		curl_setopt($curl, CURLOPT_VERBOSE, $debug);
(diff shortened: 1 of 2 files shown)


aik099:
Please do this:

1. add `protected $timeout;` property to `CurlClient` class
2. add optional `$timeout = null` parameter to `CurlClient` class constructor, that will be assigned to that `$timeout` class property
3. if `$this->timeout` is numeric, then set timeout on curl resource
4. add tests to confirm, that timeout is respected

This would avoid BC break with enforced timeout of 30 seconds.

aik099:
@ish1301 , I saw new commit, but it does only part of requested changes. I guess more commits are coming. Anyway, when you're ready to for code review just leave a message in here.

ish1301:
I can't do 2nd step, because that will force me to change ClientInterface and timeout can't be applied to other classes e.g. PHPClient.php

The only way to change timeout is `$this->api->client->timeout = 30;`

aik099:
> I can't do 2nd step, because that will force me to change ClientInterface

I haven't said, that you need to change `ClientInterface`. I said, that you need to change `CurlClient` class constructor (which isn't part of the `ClientInterface` interface) so that you can specify curl-specific timeout. The timeout should be anything present in all client interface implementing parties.

> The only way to change timeout is $this->api->client->timeout = 30;

That won't do of course.

ish1301:
Adding new parameter `sendRequest()` function returns below error, because it expects `sendRequest()` function to be exact structure as it's in ClientInterface

`Declaration must be compatible with ClientInterface->sendRequest`

ish1301:
screen shot 2017-11-13 at 3 45 34 pm

aik099:
> Adding new parameter sendRequest() function returns below error, because it expects sendRequest() function to be exact structure as it's in ClientInterface
> 
> Declaration must be compatible with ClientInterface->sendRequest

Again I don't see when I requested to change that method signature.

Please don't do things that aren't requested.

aik099:
Closing in favor of #216.

PR #162, opened 2017-10-17
API endpoint URL query string parameters

mustdobetter:
Adding support for specification of API endpoint URL query string parameters when editing an issue. Useful when, for example, you wish to prevent users from being notified of the update i.e. ?notifyUsers=false.

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -170,6 +170,21 @@ class Api
 		$this->resolutions = null;
 	}
 
+    /**
+     * Helper method to formulate a query string from supplied parameter array.
+     *
+     * @param $params Array of URL parameters: [key => value]
+     * @return string
+     */
+	protected function formulateQueryString($params)
+    {
+        $query_string = "?";
+        foreach ($params as $key => $value) {
+            $query_string .= "$key=$value&";
+        }
+        return rtrim($query_string, "&");
+    }
+
 	/**
 	 * Get fields definitions.
 	 *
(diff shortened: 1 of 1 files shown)


src/Jira/Api.php
```diff
@@ -170,6 +170,21 @@ protected function clearLocalCaches()
 		$this->resolutions = null;
 	}
 
+    /**
+     * Helper method to formulate a query string from supplied parameter array.
+     *
+     * @param $params Array of URL parameters: [key => value]
+     * @return string
+     */
+	protected function formulateQueryString($params)
```
aik099: There is PHP built-in function, that does it: http://php.net/manual/en/function.http-build-query.php

src/Jira/Api.php
```diff
@@ -696,26 +714,32 @@ public function createRemotelink(
...
+     */
 	public function api(
 		$method = self::REQUEST_GET,
 		$url,
 		$data = array(),
 		$return_as_array = false,
 		$is_file = false,
-		$debug = false
+		$debug = false,
+        $url_params = array()
```
aik099: This method already has too many params, so adding another one makes it a nightmare. Any other implementations you can suggest?

My options are:

1. having special key in `$data` array that won't match any parameter of existing in JIRA API (not necessary implemented in this library) for passing query parameters
2. use `http_build_query` directly on `editIssue` method and not add any parameters here

aik099: Implemented in the #222.

PR #167, opened 2017-12-26
force fields type to array

djalal:
fixes the following error message : 

```
Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/html/vendor/chobie/jira-api-restclient/src/Jira/Api.php on line 734
```

### DIFF
--- a/src/Jira/Issues/Walker.php
+++ b/src/Jira/Issues/Walker.php
@@ -139,8 +139,9 @@ class Walker implements \Iterator, \Countable
 	 */
 	public function push($jql, $fields = null)
 	{
+		if (is_string($fields)) $fields=[$fields];
 		$this->jql = $jql;
-		$this->fields = $fields;
+		$this->fields = is_string($fields) ? [$fields] : $fields;
 	}
 
 	/**
--- a/src/Jira/Issues/Walker.php
+++ b/src/Jira/Issues/Walker.php
@@ -141,7 +141,7 @@ class Walker implements \Iterator, \Countable
 	{
 		if (is_string($fields)) $fields=[$fields];
 		$this->jql = $jql;
-		$this->fields = is_string($fields) ? [$fields] : $fields;
+		$this->fields = is_string($fields) ? [$fields] : (array) $fields;
 	}
 
 	/**


aik099:
The `$fields` parameter of `Walker::push` method is passed to the `Api::search` which ends up as `$data` in `Api::api` and not `$this->fields` which used in `count($this->fields)` line at 734.

Anything specific about your code, that might be triggering this issue? I'm not seeing it in tests. By the way tests are massively failing after your changes. That needs to be fixed.

dereuromark:
I get the same

> count(): Parameter must be an array or an object that implements Countable in [/home/vagrant/Apps/.../vendor/chobie/jira-api-restclient/src/Jira/Api.php, line 734]

with php7.2

aik099:
@dereuromark , can you please post test code you're using or better year a PR with a failing test?

dereuromark:
Nothing fancy.
Note that it works fine in 7.1, in 7.2 throws the warnings.

$query = '...';
$walker->push($query);

as documented in your walker class, it may/must be a string (which i do):
```
	/**
	 * Pushes JQL.
	 *
	 * @param string            $jql    JQL.
	 * @param string|array|null $fields Fields.
	 *
	 * @return void
	 */
	public function push($jql, $fields = null)
	{
```

aik099:
As I've said before the `$fields` parameter value you're casting isn't `Api::fields` property value you're getting an error from.

The error comes from 734 line and in there `if ( !count($this->fields) ) {` code is found. The `$this->fields` has `null` initial value and therefore it fails. That IF needs to be altered to `!isset($this->fields)`.

aik099:
In meanwhile I've sent couple of PRs to make sure that `Warning` will appear during build so that we can verify the fix:

* #168
* #169

aik099:
I've merged above PR's, but on PHP 7.2 not seeing any warnings. Maybe we're not having a test case, when `Walker::push` method is used with string passed in 2nd argument.

aik099:
I've added a failing test in #173 and a fix in #174.

aik099:
Replaced via #174.

PR #174, opened 2018-02-08
Fixes notice about "count" function usage on PHP 7.2

aik099:
Replaces #167


dereuromark:
Will you do a new release here?

aik099:
At some point probably. Since next release is 2.0.0 with BC breaks allowed I tend to break BC to get nicer APIs. Unfortunately I don't have much time to finish that job and therefore release not coming any time soon. There are still plenty of issues to be resolved before release anyway.

You're welcome to use `dev-master` as version in `composer.json` as other people do for now.

dereuromark:
I can confirm that master is now working fine.

PR #177, opened 2018-09-26
Allow Empty Result

lsklyut:
This fixes: "Argument 1 passed to chobie//Jira//Api//Result::__construct() must be of the type array, null given, called in /var/www/site/vendor/chobie/jira-api-restclient/src/Jira/Api.php on line 748 and defined...." Potentially, there should be some sort of indication that the result is not a success

### DIFF
--- a/src/Jira/Api/Result.php
+++ b/src/Jira/Api/Result.php
@@ -70,7 +70,7 @@ class Result
 	 *
 	 * @param array $result Result.
 	 */
-	public function __construct(array $result)
+	public function __construct(array $result = array())
 	{
 		if ( isset($result['expand']) ) {
 			$this->expand = explode(',', $result['expand']);


aik099:
What API call results in `$json` variable being `null` after `json_decode` is called in `chobie\Jira\Api::api` method?

Please post:

1. actual API call, that is made with real arguments
2. a full trace
3. `$result` variable value (from `var_dump` call), that is returned by JIRA without any transformations
4. used PHP version
5. used JIRA version

Since we only calling `json_decode`, when `$result` isn't empty, then likely API results a non-json output, which is isn't good as well.

lsklyut:
This is going to take some time to get the debug information - the error is occurring only sometimes and within the library code, so I cannot get any good debug information without forking the package and adding debugging code.

lsklyut:
Also in retrospect, maybe allowing a blank response isn't a great solution - if there's not an existing approach to distinguishing a bad response, maybe we throw an ApiException on non-200 responses

aik099:
@lsklyut , the non-200 responses is a normal thing in REST API, because for example 201 means, that JIRA issue was created.

I'm in no hurry, provide actual debug info when you have time. It's better to wait some time, but to have a solid test cased based on which code needs to be modified, then modifying code without real proof, that it needs changing at all.

PR #178, opened 2018-12-06
createRemotelink: capitalize globalId properly

glensc:
By docs the JSON key is `globalId` not `globalid`. 

JIRA version at https://pycontribs.atlassian.net does not accept wrongly cased key.

- https://developer.atlassian.com/server/jira/platform/jira-rest-api-for-remote-issue-links/
- https://developer.atlassian.com/server/jira/platform/using-fields-in-remote-issue-links/


glensc:
updated.

aik099:
Merging, thanks @glensc .

glensc:
btw, jira itself is annoying, it leaves activity comment for each update even if the globalId is present and other payload is identical.

so, for my own solution, i might need to keep track of updated remote links locally or find a way to query existing remote links and update only non-matching ones.

PR #181, opened 2019-02-05
Support "scope" as part of an IssueType

danillofb:
Scope is a new key that can be found on the issuetypes array. 
Since this is not whitelisted on the acceptableKeys array, if this function is needed for anything and the project have a scope, the process dies before reaching out the end of the execution.

The existence of the scope key can be found on the api documentation: https://developer.atlassian.com/cloud/jira/platform/rest/v2/#api-api-2-issue-createmeta-get

### DIFF
--- a/src/Jira/IssueType.php
+++ b/src/Jira/IssueType.php
@@ -90,6 +90,7 @@ class IssueType
 		'name',
 		'subtask',
 		'avatarId',
+		'scope',
 	);
 
 	/**


aik099:
Merging, thanks @danillofb .

glensc:
@aik099 why not tag some 2.x release, the 1.x tag is 5 years old, and pretty useless i think.

glensc:
... but according to semver, 2.x is new major, you are allowed to break.

i rather see 2.x with some inconsistencies than using dev-master (yes i know about composer.lock, but still)

i believe that if you actually create 2.x tags, it might tract interested parties to work on the unsolved issues, or you can decide they are not important and close instead.

aik099:
I mean between 1.x and 2.x you can do a BC breaks. I don't want to release 2.x if I know upfront, that I'll be doing tasks that cause a BC break and I'll be forced to release 3.x shortly after that.

Issue #120, opened 2016-07-23
Issues with "Api::createRemotelink" method


aik099:
The `Api::createRemotelink` method (added in 2.0.0) have following problems:
* [x] it's name violates camelCase naming rules, where each word (except 1st one) in method name must be capitalized (should be `Api::createRemoteLink`)
* [x] the `globalid` parameter in API call should be named `globalId` instead (see https://docs.atlassian.com/jira/REST/latest/#api/2/issue-createOrUpdateRemoteIssueLink)
* [ ] the `globalid` and `relationship` parameters shouldn't be passed to API call when they values are empty

Related to #43

glensc:
the second point was fixed in https://github.com/chobie/jira-api-restclient/pull/178

cross it out or add link ;)

aik099:
> the second point was fixed in #178
> 
> cross it out or add link ;)

Done.

aik099:
> it's name violates camelCase naming rules, where each word (except 1st one) in method name must be capitalized (should be Api::createRemoteLink)

Done in #183.

aik099:
> the globalid and relationship parameters shouldn't be passed to API call when they values are empty

Fixed in the #231.

PR #183, opened 2019-02-13
rename createRemotelink -> createRemoteLink

glensc:
Fixes "1" from #120:

- it's name violates camelCase naming rules, where each word (except 1st one) in method name must be capitalized (should be Api::createRemoteLink)

method names are case insensitive in PHP, unless there's some magic getters/setters behind, this change is backward compatible.

cc @aik099

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Remove trailing slash from endpoint url by [@Procta].
 - Added local cache to getResolutions by [@jpastoor].
 - Renamed Api::api() parameter $return_as_json to $return_as_array by [@jpastoor].
+- Renamed `Api::createRemotelink` to `Api::createRemoteLink` by [@glensc].
 
 ### Removed
 ...
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -676,7 +676,7 @@ class Api
 	 * @return array|false
 	 * @since  2.0.0
 	 */
-	public function createRemotelink(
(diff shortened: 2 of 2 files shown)


glensc:
I mean, that the change will not break any logic as method names are case insensitive in PHP, unless some __set or __get is involved, I'm not familiar with the codebase.

glensc:
so this goes under `changes` or `added` or err?

not sure this needs changelog entry, as you said, the code was not released.

changing case of previous changelog entry seems also weird.

```
- Added `Api::createRemotelink` call for creating linking issue with remote applications by [@elmi82].
```

aik099:
This change should go under `changed`. Yeah, a bit weird theoretically, but since different people added method and changed it I see no logical conflict.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Remove trailing slash from endpoint url by [@Procta].
 - Added local cache to getResolutions by [@jpastoor].
 - Renamed Api::api() parameter $return_as_json to $return_as_array by [@jpastoor].
+- Rename createRemotelink -> createRemoteLink [@glensc].
```
aik099: Please change to:

```
Renamed `Api::createRemotelink` into `Api::createRemoteLink` by [@glensc].
```

Also you should map your GitHub username to GitHub user profile page at the bottom of the file to make your username clickable.

glensc: updated. the link was added already by my previous contribution

aik099:
Merging, thank you @glensc .

PR #184, opened 2019-02-13
Apply code style fixes

glensc:
Minor code style fixes done with help of [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer).

### DIFF
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -741,16 +741,15 @@ class Api
 				}
 			}
 
-			if ( $return_as_array ) {
+			if ($return_as_array) {
 				return $json;
 			}
-			else {
-				return new Result($json);
-			}
+
+			return new Result($json);
+
 		}
-		else {
-			return false;
-		}
+
+		return false;
 	}
 
 	/**
--- a/src/Jira/Api.php
+++ b/src/Jira/Api.php
@@ -741,12 +741,11 @@ class Api
 				}
 			}
 
-			if ($return_as_array) {
+			if ( $return_as_array ) {
 				return $json;
 			}
 
 			return new Result($json);
-
 		}
 
 		return false;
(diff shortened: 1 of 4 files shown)


glensc:
1. not directly. i was about to make changes, but noticed `is_null` in place
2. i don't use the tool. shouldn't travis complain on this?

aik099:
> 1. not directly. i was about to make changes, but noticed `is_null` in place

I see.

> 2. i don't use the tool. shouldn't travis complain on this?

It's not doing that currently. It makes sense to do only, when we fix all existing issues to avoid all new build be broken by default.

glensc:
checked `vendor/bin/phpcs`, the changes here are not related.

aik099:
Maybe can combine all CS fixes into single PR? I've just updated `composer.json` and `composer.lock` to ensure that used CS is compatible with latest PHP_CodeSniffer version.

PHP_CodeSniffer also does support auto-fixing of most, but not all issues found.

If you're using PhpStorm, then you configure it to automatically highlight problematic places.

We're not using Scrutinizer CI or otherwise it will be showing CS issues online as well.

glensc:
You can squash merge from GitHub web UI if you have enabled that in project. If that's what you mean by single commit.

glensc:
At this moment, I do not plan to make more CS fixes.

src/Jira/Api.php
```diff
@@ -741,16 +741,15 @@ public function api(
 				}
 			}
 
-			if ( $return_as_array ) {
+			if ($return_as_array) {
```
aik099: Please restore spaces, that were removed.

---

As you can see from majority of above changed IFs the condition is wrapped in spaces.

src/Jira/Api.php
```diff
@@ -741,16 +741,15 @@ public function api(
...
 			}
-			else {
-				return new Result($json);
-			}
-		}
-		else {
-			return false;
+
+			return new Result($json);
+
```
aik099: Please remove an empty line.

---

No need to add an empty line before closing `}`.

aik099:
Merging, thanks @glensc .

PR #185, opened 2019-02-13
Setup .gitattributes for better changelog merges

glensc:
While github.com still doesn't support this in their web, this has an effect on rebase/merges.

ps: gitlab.com web supports this ;)
- https://gitlab.com/gitlab-org/gitlab-ce/issues/17325


aik099:
Merging, thank you @glensc .

PR #190, opened 2019-07-03
Add getTimeTracking function.

criz:
Otherwise timetracking data is not available in issues.

### DIFF
--- a/src/Jira/Issue.php
+++ b/src/Jira/Issue.php
@@ -332,5 +332,14 @@ class Issue
 
 		return null;
 	}
-
+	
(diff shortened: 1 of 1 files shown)


aik099:
The `$issue->get('Time Tracking')` would also work, because `get` method is public.

criz:
The get function is only returning fields.

aik099:
But inside your new method you're using that `get` method as well.

criz:
haha, yeah. Sorry, I was confused with this only returning "fields"...

PR #195, opened 2020-10-29
feat: set a proxy to use for the request

### LINKED ISSUE #194: Allow set proxy to use
Set a proxy to use for the request

aalcaidep:
Closes #194.


aik099 (commented):
I see, that you're specifying proxy per API-call. Please specify a use case for such implementation.

If you don't mind, then please change implementation to allow specifying proxy upon CurlClient construction.

aik099:
Closing in favor of #216.

Issue #199, opened 2022-09-26
Deprecation warnings


dereuromark:
https://github.com/chobie/jira-api-restclient/blob/master/src/Jira/Api.php#L713

`$url` should be `$url = null` or alike to avoid
```
PHP Deprecated:  Required parameter $url follows optional parameter $method in /home/travis/build/spryker/release-app/vendor/chobie/jira-api-restclient/src/Jira/Api.php on line 711
PHP Deprecated:  Required parameter $endpoint follows optional parameter $data in /home/travis/build/spryker/release-app/vendor/chobie/jira-api-restclient/src/Jira/Api/Client/CurlClient.php on line 62
PHP Deprecated:  Required parameter $credential follows optional parameter $data in /home/travis/build/spryker/release-app/vendor/chobie/jira-api-restclient/src/Jira/Api/Client/CurlClient.php on line 62
PHP Deprecated:  Required parameter $endpoint follows optional parameter $data in /home/travis/build/spryker/release-app/vendor/chobie/jira-api-restclient/src/Jira/Api/Client/ClientInterface.php on line 53
PHP Deprecated:  Required parameter $credential follows optional parameter $data in /home/travis/build/spryker/release-app/vendor/chobie/jira-api-restclient/src/Jira/Api/Client/ClientInterface.php on line 53
```

dereuromark:
Released as https://github.com/dereuromark/jira-api-restclient/releases/tag/1.1.0

PR #200, opened 2022-09-29
Fix up signature for PHP 8+

dereuromark:
Solves https://github.com/chobie/jira-api-restclient/issues/199


dereuromark:
@aik099 I recommend changing the github config for first time contributors to allow running CI :)
The default unfortunately is not the "new people" only, but everyone on github.

dereuromark:
I also added PHPStan which was definitly needed, quite a few issues that have been detected, including the PHP 8+ issues:

```
 ------ ----------------------------------------------------------------------------------------- 
  Line   Jira/Api/Client/ClientInterface.php                                                      
 ------ ----------------------------------------------------------------------------------------- 
  57     Deprecated in PHP 8.0: Required parameter $endpoint follows optional parameter $data.    
  58     Deprecated in PHP 8.0: Required parameter $credential follows optional parameter $data.  
 ------ ----------------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------------- 
  Line   Jira/Api/Client/CurlClient.php                                                           
 ------ ----------------------------------------------------------------------------------------- 
  66     Deprecated in PHP 8.0: Required parameter $endpoint follows optional parameter $data.    
  67     Deprecated in PHP 8.0: Required parameter $credential follows optional parameter $data.  
 ------ ----------------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------------- 
  Line   Jira/Api/Client/MemcacheProxyClient.php                                                  
 ------ ----------------------------------------------------------------------------------------- 
  79     Deprecated in PHP 8.0: Required parameter $endpoint follows optional parameter $data.    
  80     Deprecated in PHP 8.0: Required parameter $credential follows optional parameter $data.  
 ------ ----------------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------------- 
  Line   Jira/Api/Client/PHPClient.php                                                            
 ------ ----------------------------------------------------------------------------------------- 
  99     Deprecated in PHP 8.0: Required parameter $endpoint follows optional parameter $data.    
  100    Deprecated in PHP 8.0: Required parameter $credential follows optional parameter $data.  
 ------ ----------------------------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Jira/Issues/Walker.php                                                                                                                             
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------- 
  152    Return type mixed of method chobie\Jira\Issues\Walker::current() is not covariant with tentative return type mixed of method Iterator::current().  
         💡 Make it covariant, or use the #[\ReturnTypeWillChange] attribute to temporarily suppress the error.                                             
  171    Return type mixed of method chobie\Jira\Issues\Walker::next() is not covariant with tentative return type void of method Iterator::next().         
         💡 Make it covariant, or use the #[\ReturnTypeWillChange] attribute to temporarily suppress the error.                                             
  182    Return type mixed of method chobie\Jira\Issues\Walker::key() is not covariant with tentative return type mixed of method Iterator::key().          
         💡 Make it covariant, or use the #[\ReturnTypeWillChange] attribute to temporarily suppress the error.                                             
  201    Return type mixed of method chobie\Jira\Issues\Walker::valid() is not covariant with tentative return type bool of method Iterator::valid().       
         💡 Make it covariant, or use the #[\ReturnTypeWillChange] attribute to temporarily suppress the error.                                             
  263    Return type mixed of method chobie\Jira\Issues\Walker::rewind() is not covariant with tentative return type void of method Iterator::rewind().     
         💡 Make it covariant, or use the #[\ReturnTypeWillChange] attribute to temporarily suppress the error.                                             
  280    Return type mixed of method chobie\Jira\Issues\Walker::count() is not covariant with tentative return type int of method Countable::count().       
         💡 Make it covariant, or use the #[\ReturnTypeWillChange] attribute to temporarily suppress the error.                                             
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------- 
```

dereuromark:
Some other issues I found that are not optional:
- Deprecations around `PHPUnit\Framework\TestCase::prophesize() is deprecated and will be removed in PHPUnit 10. Please use the trait provided by phpspec/prophecy-phpunit.`
- composer.lock file is commited, which it should not be, instead the upper and lower limit should be tested
- PHP 5.4 at this point is definitely not warranted anymore, maybe upgrade to 7.3 or better 7.4 now.
- it also suffices to only test 7.4 + 8.2 for example (min/max)

src/Jira/Issues/Walker.php
```diff
@@ -289,19 +294,18 @@ public function count()
...
-		if ( is_callable($callable) ) {
-			$this->callback = $callable;
-		}
-		else {
-			throw new \Exception('passed argument is not callable');
+		if ( !is_callable($callable) ) {
+            throw new \Exception('passed argument is not callable');
 		}
+
+        $this->callback = $callable;
```
aik099, line 308: Code indentation issue. Likely due to mixed tabs/spaces usage.

src/Jira/Issues/Walker.php
```diff
@@ -289,19 +294,18 @@ public function count()
 	/**
 	 * Sets callable.
 	 *
-	 * @param callable $callable Callable.
+	 * @param mixed $callable Callable.
```
aik099, line 297: Why this change? The method accepts only callables. Changing signature would better express this, however, that would also be a BC break.

src/Jira/Issues/Walker.php
```diff
@@ -277,6 +281,7 @@ public function rewind()
 	 * @return integer The custom count as an integer.
 	 * @link   http://php.net/manual/en/countable.count.php
 	 */
+    #[\ReturnTypeWillChange]
```
aik099, line 284: This kind of stuff would result in an exception/fatal error on PHP versions prior to PHP 8.0. Definitely remove.

src/Jira/Issues/Walker.php
```diff
@@ -260,6 +263,7 @@ public function valid()
 	 * @return void Any returned value is ignored.
 	 * @link   http://php.net/manual/en/iterator.rewind.php
 	 */
+    #[\ReturnTypeWillChange]
```
aik099, line 266: This kind of stuff would result in an exception/fatal error on PHP versions prior to PHP 8.0. Definitely remove.

aik099 (changes requested):
Please remove all changes in this PR, that:

- cause any test failures
- doesn't really solve an issue of #199

If you would like to propose some other changes, like test suite improvement, supported PHP version change, or anything else, then please do so via an issue creation. Once, in the issue, we agree, that project will benefit from these changes, then and only then, you can proceed with the development.

This way you can save both developer time (your time) and reviewer time (my time).

P.S.
The PR must be on topic.

src/Jira/Issues/Walker.php
dereuromark: An editorconfig file would prevent this, I added the missing file.

src/Jira/Issues/Walker.php
dereuromark: No, you are wrong on both counts
The code clearly accepts anything else besides callable and throws an exeption then
If you wouldnt want to accept those, the guard clause shouldnt be in there.
It is also not a BC break to fix up documentation on those
PHPStan correctly calls out the code for this.

dereuromark: We could fake it by saying callable|null or alike
But it would still not be quite the truth.

src/Jira/Issues/Walker.php
dereuromark: It does not, as it is a comment here. See the CI results of those versions.
But it is needed in higher PHP versions
As PHPStan correctly points out.

src/Jira/Issues/Walker.php
```diff
@@ -117,7 +117,7 @@ class Walker implements \Iterator, \Countable
 	/**
 	 * Creates walker instance.
 	 *
-	 * @param Api          $api      API.
+	 * @param Api		  $api	  API.
```
dereuromark, line 120: Darn, I wasnt exoecting the anti pattern of code alignment to be used, I guess I have to fix that up later by reverthing those changes,

dereuromark:
PHPStan isnt easily runnable on PHP < 7.1, I would recommend dropping as as written above, since those are long EOL anyway.

dereuromark:
Regarding saving time: It would save both our time if the github config was to be adjusted to the recommended setting.

dereuromark:
Refs https://github.com/Riverline/jira-api-restclient/pull/1

src/Jira/Api/Client/CurlClient.php
```diff
@@ -146,7 +146,7 @@ public function sendRequest(
 		}
 
 		// @codeCoverageIgnoreStart
-		if ( is_null($response) ) {
+		if ( $response === false ) {
```
aik099, line 149: Why you've changed this?

**This change should be covered by tests.**

src/Jira/Api/Client/PHPClient.php
```diff
@@ -201,11 +201,13 @@ protected function doSendRequest($url, array $context)
 	{
 		$this->_lastErrorMessage = '';
 
-		set_error_handler(array($this, 'errorHandler'));
+		/** @var callable $callable */
+		$callable = array($this, 'errorHandler');
+		set_error_handler($callable);
```
aik099, line 206: Why you've changed this? The original code version looks working to me.

src/Jira/Api/Client/PHPClient.php
```diff
@@ -201,11 +201,13 @@ protected function doSendRequest($url, array $context)
...
 		$response = file_get_contents($url, false, stream_context_create($context));
 		restore_error_handler();
 
-		if ( isset($http_response_header) ) {
+		if ( $http_response_header ) {
```
aik099, line 210: The changed code version expects, that `$http_response_header` would always be defined. Is this the case?

**This change should be covered by tests.**

src/Jira/Issues/Walker.php
aik099: Please:

1. revert the `DocBlock`, so it would use the `callable` data type again
2. add `callable` typehint to the `setDelegate` method (this method)
3. remove the guard clause

---

I think we can safely do this now without causing a BC break because PHP 5.3 is no longer supported.

src/Jira/Issues/Walker.php
aik099: We'll see once the tests are fixed.

src/Jira/Issues/Walker.php
aik099: We'll see once the tests are fixed.

.github/workflows/tests.yml
```diff
@@ -20,28 +20,26 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        php: [ '5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1' ]
```
aik099, line 23: If I remember correctly I haven't agreed to remove support for any of the PHP versions.

Please revert and keep PHP 7.x tool usage (which PHPStan is) locked to these PHP versions.

.github/workflows/tests.yml
```diff
@@ -60,4 +58,7 @@ jobs:
 
       - name: Run tests
         run: |
-          vendor/bin/phpunit -v
+          vendor/bin/phpunit
+
+      - name: Run phpstan
+        run: composer stan-setup && vendor/bin/phpstan analyse --error-format=github
```
aik099, line 64: Please restrict to PHP versions, where PHPStan can be run.

composer.json
```diff
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0"
+        "php": ">=7.3.0"
```
aik099, line 15: Please revert.

---

See above.

composer.json
```diff
@@ -22,17 +22,26 @@
...
     "autoload-dev": {
         "psr-4": {
-            "Tests\\chobie\\": "tests"
+            "Tests\\chobie\\": "tests/"
         }
     },
+	"scripts": {
+		"stan": "phpstan analyse",
+		"stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:^1.0.0 && mv composer.backup composer.json"
+	},
```
aik099, line 36: Please remove. Instead, do this inside a build workflow.

Theoretically, the `phpstan analyse` command should fail, because PHPStan is installed locally and therefore can only be accessed via `vendor/bin/phpstan`.

composer.json
```diff
@@ -22,17 +22,26 @@
...
     "extra": {
         "branch-alias": {
             "dev-master": "2.0-dev"
         }
+    },
+    "config": {
+        "platform": {
+            "php": "7.3"
+        }
     }
```
aik099: Please rollback.

Such an approach would prevent installing dependencies for other PHP versions.

phpunit.xml.dist
```diff
@@ -11,8 +11,6 @@
          stopOnFailure="false"
          stopOnIncomplete="false"
          stopOnSkipped="false"
-         syntaxCheck="true"
-         strict="true"
```
aik099, line 15: As I've said earlier (I do remember my all comments even though you seem to have force pushed, which erased my previous review results), we need to replace `strict="true"` with the corresponding `beStrict...` options instead of just switching to non-strict mode.

README.md
```diff
@@ -1,6 +1,11 @@
 # Jira REST API Client
 
 [![CI](https://github.com/chobie/jira-api-restclient/actions/workflows/tests.yml/badge.svg)](https://github.com/chobie/jira-api-restclient/actions/workflows/tests.yml)
+[![PHPStan](https://img.shields.io/badge/PHPStan-level%206-brightgreen.svg?style=flat)](https://phpstan.org/)
```
aik099, line 4: Is this a static shield?

.github/workflows/tests.yml
```diff
@@ -60,4 +58,7 @@ jobs:
 
       - name: Run tests
         run: |
-          vendor/bin/phpunit -v
+          vendor/bin/phpunit
```
aik099, line 61: Why you've removed the `-v` option?

.github/workflows/tests.yml
```diff
@@ -20,28 +20,26 @@ jobs:
...
           php-version: "${{ matrix.php }}"
           tools: composer
-          # PHP 7.1 development web server segfaults if timezone not set.
+          # PHP 7.x development web server segfaults if timezone not set.
           ini-values: date.timezone=Europe/Paris, error_reporting=-1, display_errors=On
 
-      - name: Configure for PHP >= 7.1
-        if: "${{ matrix.php >= '7.1' }}"
+      - name: Configure for PHP >= 7.3
+        if: "${{ matrix.php >= '7.3' }}"
```
aik099, line 42: Please rollback.

.github/workflows/tests.yml
```diff
@@ -20,28 +20,26 @@ jobs:
...
+        php: [ '7.3', '7.4', '8.0', '8.1' ]
       fail-fast: false
     env:
       REPO_URL: http://localhost:8002/
 
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-        with:
-          fetch-depth: 2
```
aik099, line 32: Why this was removed?

aik099 (changes requested):
The goal of any collaboration is to get to an agreement. Observing the changes you've done in this PR can't be called an agreement.

If you want this PR to be merged I recommend listening to and discussing the suggestions I'm during a review.

aik099:
> Regarding saving time: It would save both our time if the github config was to be adjusted to the recommended setting.

I have no idea why GitHub is asking me to approve every commit you make on this PR. If you have any idea why then please share it with me.

Anyway since a typo PR of yours was now merged I'd expect GitHub to treat you as a non-first-time contributor and will auto-run builds on your commits.

src/Jira/Api/Client/PHPClient.php
dereuromark: for PHPStan, as this is a deprecated callable

src/Jira/Api/Client/PHPClient.php
dereuromark: PHPStan knows this can never be null, only empty.

.github/workflows/tests.yml
dereuromark: not needed with v2+

.github/workflows/tests.yml
dereuromark: not needed, is it? already quite verbose

README.md
dereuromark: yes

phpunit.xml.dist
dereuromark: does that make a difference?

dereuromark:
```
	public function testSetDelegateError()
	{
		$this->expectException('\Exception');
		$this->expectExceptionMessage('passed argument is not callable');

		$walker = $this->createWalker();
		$walker->setDelegate('not a callable');
	}
```
Adding a typehint is more radical than just fixing the documentation. I am not a fan of that, it would also break above test (which then we need to remove).

src/Jira/Api/Client/PHPClient.php
aik099: I see. Next time you can post links such as https://php.watch/versions/8.2/partially-supported-callable-deprecation to same me some time figuring this for myself.

src/Jira/Api/Client/PHPClient.php
aik099: Oh, I wouldn't blindly trust any static code analyzer. Please back this change with the corresponding tests.

.github/workflows/tests.yml
aik099: I see now. Thank you.

src/Jira/Api/Client/PHPClient.php
dereuromark: I do in this case. Those are correct, if you also look into source code and PHP docs.
The code itself doesnt make sense as it was.

.github/workflows/tests.yml
aik099: The `quite verbose` term can't be measured. If you'd like this option to be removed please show the comparison of both PHPUnit runs (with `-v` and without it) to see what we've loosing by removing this option.

README.md
aik099: Thank you.

phpunit.xml.dist
aik099: The `strict` option was split out into the several`beStrict...` options and some of them are enabled by default. This way you have to figure out:

1. which of `beStrict...` options are disabled by default
2. would benefit we'll gain by enabling them
3. enable appropriate options

I'll trust you on this.

aik099:
> True, the lock file should be completely removed, as I showed in the PHP8.2 check
> but I reverted the version locking for now.

You're not following, are you? No dropping supported PHP versions. No removing `composer.lock`.

dereuromark:
> If you want this PR to be merged I recommend listening to and discussing the suggestions I'm doing in this review.

I want to be clear to you here.
I am putting my work + free time here into getting your project maintained and on a quality level again.
It was clearly not very maintained so far. This is blocking us now on a critical timeframe.
If you are not interested in an actual collaboration and moving forward here, I will have to fork and create own maintenance on this within the next 48h to unblock us.
I am not interested in a long discussion on some of the critical key components, including proper stan check and up to date test matrix. I have been using those for years on 1000+ repos and I do know when I can trust them. It would be wise for you to do the same.
PHP 8.2 shows that here are even hidden "gotchas" like non sensical null checks, and even property assignment issues ($subTask vs $subtask) that will blow up in PHP 8.2.
On other issues you saw that I am open for discussion, where it makes sense. I do need to push back however on reverting some important improvements.

As for versions: Feel free to add again whats possible, but it is clearly not needed and shouldnt be < 7.3 ever at this point.
I dont want to spend time here to try to make things maintainble going that far back while keeping PHPStan support on.

So again: We need to move forward here with both our times not wasted much more.
Please let me know if you prefer to have it maintained collaboratively - or want me to start maintaining a fork.

Apart from this I recommend you to try to
- update PHPStan to recommended level 7/8 - 8 is nullable safety which is of course nice to have
- remove lock file as this is a clear anti pattern in library dev

src/Jira/Api/Client/PHPClient.php
aik099: I admire the energy you're putting into defending the static code analyzer. I would love to see that redirected into test writing. :) Not joking.

aik099:
Thanks for bringing some clarity on your motivation here.

My vision is: do the minimal changes, that are necessary to make things happen. That includes:

* if it works - don't change it
* whoever needs a feature added I'm up to it (as a reviewer) as long as we agree on the implementation
* bugfixes are always welcome
* new PHP version compatibility - why not, but not dropping support for other PHP versions (I don't care if they're not used by famous people in the community)

Time is a precious resource, and I don't want to spend it on things nobody needs.

To summarize: if you're in a hurry, then please do what you see fit in your fork. When you have time, then please come back with PRs you are willing to share.

Thank you.

PR #205, opened 2024-10-10
Make code PHP8.1 compatible

royarisse:
Fix: Remove optional arguments before required arguments.
Fix: Return types for Walker.

### DIFF
--- a/src/Jira/Issues/Walker.php
+++ b/src/Jira/Issues/Walker.php
@@ -149,7 +149,7 @@ class Walker implements \Iterator, \Countable
 	 * @return mixed Can return any type.
 	 * @link   http://php.net/manual/en/iterator.current.php
 	 */
-	public function current()
+	public function current(): mixed
 	{
 		if ( is_callable($this->callback) ) {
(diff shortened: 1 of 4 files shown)


src/Jira/Issues/Walker.php
```diff
@@ -149,7 +149,7 @@ public function push($jql, $fields = null)
 	 * @return mixed Can return any type.
 	 * @link   http://php.net/manual/en/iterator.current.php
 	 */
-	public function current()
+	public function current(): mixed
```
aik099, line 152: I understand that in PHP 8.1 (or even in PHP 8.0) the `Iterator` interface method signatures were changed. Due to that change, I see no way to have the same iterator class declaration, that will satisfy every supported PHP version (5.x, 7.x, 8.x).

Do you see any way to make this work?

P.S.
This code changes causes lot of builds to fail.

aik099 (changes requested):

aik099:
All, except for the iterator-related changes was done in the #226.

PR #250, opened 2026-03-01
2.x php8 compat

alex-dna:
Hi.
We found that using the module with PHP 8 triggered a few deprecation warnings, so this MR aims at correcting them.
This is also bumping the minimum PHP version to 7.1. Not sure if that something you would want to do.


.github/workflows/tests.yml
aik099: Please do this:

1. delete the `phpunit.xml.dist`, that currently holds PHPUnit 5.x and PHPUnit 6.x configuration
2. rename the `phpunit9.xml.dist` into the `phpunit.xml.dist`
3. for the PHP 7.4 and PHP 8.0 builds (selected lines) use `phpunit.xml.dist` instead of the `phpunit9.xml.dist`

aik099: Please drop the support for PHP 7.1, PHP 7.2 and PHP 7.3 as well.

composer.json
```diff
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.6.0"
+        "php": ">=7.1.0"
```
aik099: Please do this:

1. change the constraint to the `>=7.4.0`.
2. rebuild the `composer.lock` file (you need to run `composer update` using PHP 7.4 executable)

aik099 (changes requested):
I agree with your proposal. Let's do a few more changes to prepare PR for merging.

Please also add a line in the `CHANGELOG.md` file (1st line with a change + 2nd line at the file bottom for PR author link to work if you're a 1st time contributor in this repository).

.github/workflows/tests.yml
```diff
@@ -60,11 +50,6 @@ jobs:
           # PHP 7.1 development web server segfaults if timezone not set.
           ini-values: date.timezone=Europe/Paris, error_reporting=-1, display_errors=On
 
-      - name: Configure for PHP >= 7.1
-        if: "${{ matrix.php >= '7.1' }}"
-        run: |
-          composer require --no-update --dev symfony/error-handler "^4.4 || ^5.0"
-
```
aik099, line 67: Why this was removed?

Since the `IF` is now passing for all used PHP versions, we can do this `composer require ...` command on every build.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - The `$params` argument of the `Api::getTransitions` method is now optional by [@aik099] (#244).
 - The `$start_at = 0` argument of `Api::search` method was renamed into the `$next_page_token = null` to reflect underlying Jira API changes by [@aik099] (#248).
 - Improved performance of `Result::getIssuesCount` method by [@aik099] (#248).
+- Drop support for php >7.4.0 [@alex-dna] (#250).
```
aik099: Please adjust like this:

```md
- Drop support for PHP <7.4.0 by [@alex-dna] (#250).
```

Proposed changes:
* the comparison sign was inverted
* the `php` keyword was lowercased
* the ` by ` was missing

aik099 (changes requested):
Almost there.

.github/workflows/tests.yml
alex-dna: Sorry, my bad, I mis-read the sign.

.github/workflows/tests.yml
```diff
@@ -59,7 +49,7 @@ jobs:
           tools: composer
           # PHP 7.1 development web server segfaults if timezone not set.
           ini-values: date.timezone=Europe/Paris, error_reporting=-1, display_errors=On
-
+
```
aik099: Please remove this trailing whitespace.

P.S.
It was likely added while you're removing and adding back the same yaml file fragment.

aik099 (commented):
Almost there. Everything else looks solid.

aik099 (approved):

aik099:
Merging. Thank you @alex-dna.

If you have time to spend on this project, that I'll be happy to review your PRs.

